PR #18, opened 2025-10-06
Creating pagination view model with dummy claims

jackheslop96:
Adding results table and pagination to landing page

screencapture-localhost-8080-2025-10-06-13_20_53

## Checklist

### DIFF
--- a/src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/PaginationResults.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/PaginationResults.java
@@ -1,0 +1,46 @@
+package uk.gov.justice.laa.amend.claim.viewmodels;
+
+import lombok.Getter;
+
+/** View model for handling pagination results. */
+@Getter
+public class PaginationResults {
+    private int count;
+    private int from;
+    private int to;
+    private String text;
+
+    /**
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/PaginationResults.java
```diff
@@ -0,0 +1,46 @@
+package uk.gov.justice.laa.amend.claim.viewmodels;
+
+import lombok.Getter;
+
+/** View model for handling pagination results. */
+@Getter
```
sumachalla39, line 6: We might need to use same objects used by data-claims-api?
or we will need to add some translator/mapper between claims-api response object (in this case ClaimResultSet)?

jackheslop96: @sumachalla39 yep agreed, the claim model is really just a placeholder at the moment to populate the table and pagination components.

sumachalla39: May be using separate mapper to decouple FE is better.. We can think about this once we have api client to execute calls as separate task

sumachalla39 (approved):

PR #56, opened 2025-11-05
BC-233: Storing a claim in redis using spring-session-data-redis

jackheslop96:
## BC-233

[Link to story](https://dsdmoj.atlassian.net/browse/BC-233)

Storing a claim in redis using spring-session-data-redis

Claim gets saved in redis as attribute `claimId` at key `spring:session:sessions:<sessionId>`

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
```diff
@@ -26,6 +29,8 @@ public String onPageLoad(
         ClaimResponse claimResponse = claimService.getClaim(submissionId, claimId);
         Claim claim = claimResultMapper.mapToClaim(claimResponse);
 
+        cacheRepository.set(claimId, claim);
```
sumachalla39: can this logic go inside claimService? when it is null, we can make api call to re fetch the claims. 

```
getClaim (...) {
    claimResponse = claimsApiClient.searchClaims(...)
    if (claim == null) {
        ClaimResponse claimResponse = claimsApiClient.searchClaims(....) ;
        claim = claimResultMapper.mapToClaim(claimResponse);
        cacheRepository.set(claimId, claim);
    }
    }

```
Also minor suggestion, you can remove cache repository and use annotation on top of getClaims method of service
@Cacheable(value = "claims", key = "#claimId")

sumachalla39 (commented):
@jackheslop96 what happens when redis docker is not running? will the application runs? Checking, if it causes issues without redis instance on dev environment after merge?

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
jackheslop96: This line is outside the scope of this ticket to be honest it was more just for proof of concept. I think there are discussions to be had around whether we want to cache both the submitted claim (i.e. the original claim) as well as the claim with the assessed values. Or if we want the cached version to be the amendable one and the submitted one always retrieved from the API.

jackheslop96:
> @jackheslop96 what happens when redis docker is not running? will the application runs? Checking, if it causes issues without redis instance on dev environment after merge?

Yes there is a dependency on redis now. We will need to complete [BC-242](https://dsdmoj.atlassian.net/browse/BC-242) before this gets merged.

[BC-242]: https://dsdmoj.atlassian.net/browse/BC-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/main/java/uk/gov/justice/laa/amend/claim/config/CacheConfig.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.laa.amend.claim.config;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.session.data.redis.config.annotation.web.http.EnableRedisHttpSession;
+
+@Configuration
```
jackheslop96: This can be removed. Redis will then pull the value from `spring.session.timeout`

BenMillar-MOJ (approved):
LGTM, thanks for your work on this!

PR #63, opened 2025-11-12
BC-243: Small refactor

jackheslop96:
## Refactoring the cost request parameter to be an enum type

[Link to story](https://dsdmoj.atlassian.net/browse/BC-243)

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/controllers/ChangeMonetaryValueController.java
```diff
@@ -101,26 +93,26 @@ public String postMonetaryValue(
...
-        WAITING_COSTS, Assessment::getNetWaitingCostsAmount
+    private static final Map<Cost, Function<Assessment, BigDecimal>> GETTERS = Map.of(
+        Cost.PROFIT_COSTS, Assessment::getNetProfitCostsAmount,
+        Cost.DISBURSEMENTS, Assessment::getDisbursementAmount,
+        Cost.DISBURSEMENTS_VAT, Assessment::getDisbursementVatAmount,
+        Cost.COUNSEL_COSTS, Assessment::getNetCostOfCounselAmount,
+        Cost.DETENTION_TRAVEL_AND_WAITING_COSTS, Assessment::getTravelAndWaitingCostsAmount,
+        Cost.JR_FORM_FILLING_COSTS, Assessment::getJrFormFillingAmount,
+        Cost.TRAVEL_COSTS, Assessment::getNetTravelCostsAmount,
+        Cost.WAITING_COSTS, Assessment::getNetWaitingCostsAmount
```
BenMillar-MOJ: Should these mappers be part of the controller or part of the Enum, you could define your Enum as:
```
private final String path;
private final Function<Assessment, BigDecimal> getter;
private final BiConsumer<Assessment, BigDecimal> setter;

PROFIT_COSTS("profit-costs", Assessment::getNetProfitCostsAmount, Assessment::setNetProfitCostsAmount),
etc.
```

Then these mappers can be removed

src/main/java/uk/gov/justice/laa/amend/claim/models/Cost.java
```diff
@@ -0,0 +1,38 @@
...
+            case "profit-costs" -> PROFIT_COSTS;
+            case "disbursements" -> DISBURSEMENTS;
+            case "disbursements-vat" -> DISBURSEMENTS_VAT;
+            case "counsel-costs" -> COUNSEL_COSTS;
+            case "detention-travel-and-waiting-costs" -> DETENTION_TRAVEL_AND_WAITING_COSTS;
+            case "jr-form-filling-costs" -> JR_FORM_FILLING_COSTS;
+            case "travel-costs" -> TRAVEL_COSTS;
+            case "waiting-costs" -> WAITING_COSTS;
+            default -> throw new IllegalArgumentException(String.format("Unknown cost value: %s", value));
+        };
```
BenMillar-MOJ: You are redefining your enum here, meaning if you were to add a new value you would need to add to this switch case. Would it be cleaner to do something like?

```java
public static Cost fromValue(String value) {
    for (Cost c : values()) {
        if (c.path.equals(path)) {
            return c;
        }
    }
    throw new IllegalArgumentException("Unknown cost value: %s", value);
}
```

jackheslop96: Good shout 👍

BenMillar-MOJ (approved):
LGTM, thanks for this!

PR #101, opened 2025-12-03
BC-244 claim assessment view Page

sumachalla39:
## What is this PR?

[Link to story](https://dsdmoj.atlassian.net/browse/BC-244)

Added Banner with last assessed name and details
## Checklist

image


src/main/java/uk/gov/justice/laa/amend/claim/mappers/AssessmentMapper.java
```diff
@@ -54,6 +59,46 @@ public interface AssessmentMapper {
...
+    @Mapping(target = "waitingCosts.assessed", source = "netWaitingCostsAmount")
+    CrimeClaimDetails mapToCrimeClaim(AssessmentGet assessmentGet, @MappingTarget CrimeClaimDetails claimDetails);
+
+    default ClaimDetails mapAssessmentToClaimDetails(AssessmentGet assessmentGet, @MappingTarget ClaimDetails claimDetails) {
+        if (assessmentGet == null || claimDetails == null) {
+            throw new IllegalArgumentException("AssessmentGet and ClaimDetails must be non-null");
+        }
+        return claimDetails instanceof CrimeClaimDetails
+                ? mapToCrimeClaim(assessmentGet, (CrimeClaimDetails) claimDetails)
+                : mapToCivilClaim(assessmentGet, (CivilClaimDetails) claimDetails);
```
jackheslop96: Pretty minor but could make use of the `getIsCrimeClaim` boolean here

src/main/java/uk/gov/justice/laa/amend/claim/models/User.java
```diff
@@ -0,0 +1,12 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class User {
+    private String id;
+    private String displayName;
+    private String mail;
```
jackheslop96: Again very minor but should this be `email`?

src/main/resources/templates/claim-summary.html
```diff
@@ -12,6 +12,24 @@
...
+                                    <h3 class="govuk-notification-banner__heading">
+                                        This claim has been assessed
+                                    </h3>
+                                    <p class="govuk-body"
+                                       th:text="#{claimSummary.lastAssessmentText(${user?.displayName}, ${claim.lastAssessedDate}, ${claim.lastAssessedTime})}">
+                                    </p>
+                                    <p th:if="${claim.lastAssessmentOutcomeKey}" th:text="#{${'claimSummary.' + claim.lastAssessmentOutcomeKey}}"></p>
+                                </div>
+                            </div>
+                        </div>
```
jackheslop96: Consider using https://design-patterns.service.justice.gov.uk/components/alert/ . This then removes the need for any custom stylings.

src/main/java/uk/gov/justice/laa/amend/claim/utils/DateUtils.java
```diff
@@ -36,4 +39,12 @@ public static String toSubmissionPeriod(String month, String year) {
...
         }
     }
+
+    public static String getDateValue(OffsetDateTime localDate) {
+        return localDate != null ? localDate.format(DateTimeFormatter.ofPattern(DEFAULT_DATE_FORMAT)) : null;
+    }
+
+    public static String getTimeValue(OffsetDateTime localDate) {
+        return localDate != null ? localDate.format(DateTimeFormatter.ofPattern(DEFAULT_TIME_FORMAT)) : null;
+    }
```
jackheslop96: Could possibly look to refactor this so we don't need to split out the date and time. See `displayDateTimeValue` for an example (could convert the `lastAssessedDate` to a `LocalDateTime` in the mapper, similar to what we do with the `submittedDate`)

jackheslop96: I have a feeling where I've put `"dd MMMM yyyy 'at' hh:mm:ss"` it should actually be HH instead of hh as you have done, to use the 24 hour clock. If that's something you'd be able to update and unit test 🙏

src/main/java/uk/gov/justice/laa/amend/claim/mappers/AssessmentMapper.java
```diff
@@ -54,6 +59,46 @@ public interface AssessmentMapper {
     @Mapping(target = "netWaitingCostsAmount", expression = "java(mapNetWaitingCostsAmount(claim))")
     AssessmentPost mapCrimeClaimToAssessment(CrimeClaimDetails claim, @Context String userId);
 
+
+    @BeanMapping(unmappedTargetPolicy = ReportingPolicy.IGNORE)
+    @Mapping(target = "vatClaimed.assessed", source = "isVatApplicable")
```
jackheslop96, line 64: I wonder if there is a need for the `assessed` object in the ClaimField? If we were to pop these values in the `amended` field would it break anything? I'm just thinking in terms of the review and amend page. Could it potentially get quite fiddly to populate the amended column if the values are potentially coming from two different places?

src/main/java/uk/gov/justice/laa/amend/claim/service/UserRetrievalService.java
```diff
@@ -0,0 +1,52 @@
...
+        return webClientBuilder.baseUrl("https://graph.microsoft.com/v1.0")
+                .build()
+                .get()
+                .uri("/users/{upn}", upn)
+                .header(HttpHeaders.AUTHORIZATION, "Bearer " + token)
+                .retrieve()
+                .bodyToMono(User.class);
+    }
+
+    public User getGraphUser(Authentication authentication, String userId) {
```
jackheslop96: Just wondering if we need some kind of fallback here? For instance, what if the case worker who made the assessment has left the MoJ and their ID no longer exists on the tenant? If the lookup fails can we default to displaying their user ID? Happy to discuss with the wider team.

src/main/java/uk/gov/justice/laa/amend/claim/mappers/AssessmentMapper.java
sumachalla39: Good catch
Just addressed this, I am wondering since we have 3 types of area of Laws (CRIME, CIVIL, MEDIATION).. 
Currently we are dealing with only 2 area of laws (CRIME, CIVIL). So rather than using Crime or not, I have used area of law.

src/main/java/uk/gov/justice/laa/amend/claim/models/User.java
sumachalla39: This is named due to the way graphapi is returning the json.. Response has mail instead of email
We are not using this attribute, so I have removed this for now. The future approach might be changed to return displayName as part of claims from Silas, instead of making call to graphAPI call. But still needs discussions on  this.

src/main/resources/templates/claim-summary.html
sumachalla39: Good one, I have addressed this

src/main/java/uk/gov/justice/laa/amend/claim/mappers/AssessmentMapper.java
```diff
@@ -91,12 +95,14 @@ public interface AssessmentMapper {
...
             throw new IllegalArgumentException("AssessmentGet and ClaimDetails must be non-null");
         }
-        return claimDetails instanceof CrimeClaimDetails
-                ? mapToCrimeClaim(assessmentGet, (CrimeClaimDetails) claimDetails)
-                : mapToCivilClaim(assessmentGet, (CivilClaimDetails) claimDetails);
+        return switch (claimDetails.getAreaOfLaw()) {
+            case "CRIME_LOWER" -> mapToCrimeClaim(assessmentGet, (CrimeClaimDetails) claimDetails);
+            case "LEGAL_HELP", "MEDIATION" -> mapToCivilClaim(assessmentGet, (CivilClaimDetails) claimDetails);
+            default -> throw new IllegalStateException("Unexpected value: " + claimDetails.getAreaOfLaw());
+        };
```
jackheslop96: This feels like a duplication of business logic that already happens in the claim mapper? I think the `getIsCrimeClaim()` adequately fits our needs here

src/main/java/uk/gov/justice/laa/amend/claim/utils/DateUtils.java
sumachalla39: I have updated to use same for submitteDate and assessmentdate.
I have modified to get the label 'at' from message.properties to use uniform approach of localization.

src/main/java/uk/gov/justice/laa/amend/claim/utils/ThymeleafUtils.java
```diff
@@ -142,4 +157,16 @@ public boolean displayClaimField(ClaimField claimField) {
...
+    public String getLastAssessmentAlertMessage(OffsetDateTime value, String user) {
+        if (value == null || user == null) {
+            return getMessage("service.noData");
+        }
+        return getMessage("claimSummary.lastAssessmentText", user, displayFormattedFullDateValue(value.toLocalDateTime()));
+    }
+
+    public String displayFormattedFullDateValue(LocalDateTime value) {
+        return MessageFormat.format("{0} {1} {2}", getDateValue(value), getMessage("time.at"), getTimeValue(value));
+    }
```
jackheslop96, line 170: If we want to go about it this way can I suggest defining a message key `...={0} at {1}`

This would make it more robust to translations (if that is ever introduced)

src/main/java/uk/gov/justice/laa/amend/claim/mappers/AssessmentMapper.java
sumachalla39: this is new attribute - assessed is created newly in claimField
review &amend page is using amended of ClaimField object.
Happy to have discussion

src/main/java/uk/gov/justice/laa/amend/claim/service/UserRetrievalService.java
sumachalla39: Required to be discussed with wider team. As per the ticket, it didn't mention about fall back mechanism. I would wait until it actually works to get username
User X created assignment, when its viewed by User Y whether it can make a call?
For now I have made it to return null. Once experimented on dev, I am going to add a ticket to discuss on this.

jackheslop96: Cool. Happy to mark this comment as resolved 👍

src/main/java/uk/gov/justice/laa/amend/claim/mappers/AssessmentMapper.java
sumachalla39: IsCrime will only tells you whether the areaOfLaw is Crime or Civil, but there is another case MEDIATION. Improves readability on what we are using for MEDIATION

happy to discuss further

jackheslop96: That gets handled here - https://github.com/ministryofjustice/laa-amend-a-claim/blob/1605e08bcaeb7edd87adff09de515a681ab340c5/src/main/java/uk/gov/justice/laa/amend/claim/mappers/ClaimMapper.java#L94-L97

jackheslop96 (approved):

PR #103, opened 2025-12-04
BC-296 - Disabling assessment outcome for non-escaped cases

jackheslop96:
## BC-296

[Link to story](https://dsdmoj.atlassian.net/browse/BC-296)

- Using a HandlerInterceptor to:
  - check the session has a claim for the given claim ID
  - check the claim is escaped
- Disabling assessment outcome for non-escaped cases

## Checklist

### DIFF
--- a/src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
@@ -7,6 +7,7 @@ import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.server.ResponseStatusException;
 import uk.gov.justice.laa.amend.claim.service.ClaimService;
 
@@ -15,7 +16,6 @@ import uk.gov.justice.laa.amend.claim.service.ClaimService;
 public class ClaimSummaryController {
 
     private final ClaimService claimService;
-
 
     @GetMapping("/submissions/{submissionId}/claims/{claimId}")
     public String onPageLoad(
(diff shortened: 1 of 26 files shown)


src/main/java/uk/gov/justice/laa/amend/claim/interceptors/ClaimInterceptor.java
```diff
@@ -0,0 +1,54 @@
...
+        if (claim.getEscaped() == null || !claim.getEscaped()) {
+            return error(response, request, "Claim is not an escape case");
+        }
+
+        request.setAttribute(claimId, claim);
+
+        return true;
+    }
+
+    private boolean error(HttpServletResponse response, HttpServletRequest request, String message) throws IOException {
```
sumachalla39, line 52: If there is no claim in the session, is it better to send to ClaimSummary page to reload?

jackheslop96: I think eventually we'll need to redirect to this page in that scenario, so I've just kept it as a generic error for now.

Screenshot 2025-12-04 at 16 49 36

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
```diff
@@ -43,4 +44,12 @@ public String onPageLoad(
         model.addAttribute("claim", claimDetails.toViewModel());
         return "claim-summary";
     }
+
+    @PostMapping("/submissions/{submissionId}/claims/{claimId}")
```
sumachalla39, line 48: Trying to understand, why can't you use same get url as earlier? Why do you need new post method, if you are just redirecting to get assessmentOutcome page?

jackheslop96: Yes this is a fiddly one. The `disabled` and `aria-disabled=true` attributes do not have the desired effect on a link as it is still clickable, so needs to be a button

src/main/java/uk/gov/justice/laa/amend/claim/interceptors/ClaimInterceptor.java
jackheslop96: Addressed 🙂

sumachalla39 (approved):

PR #108, opened 2025-12-08
BC-370: Refactor CI-CD pipeline to use shared workflow and build artefacts

BenMillar-MOJ:
## What is this PR?

Modifies the CI-CD pipeline to use shared workflow and build artefacts.

Feature branch pipelines build, test, scan and deploy a container image to the development environment, tests do not need to pass before being deployed to allow for developers to manually test changes and gain fast feedback.

The main branch pipeline build, test, scan and deploy a container image to the UAT environment.

## Feature branch pipeline
Screenshot 2025-12-15 at 00 14 43

## Main branch pipeline
Screenshot 2025-12-15 at 00 23 35

## Checklist


.github/workflows/build.yml
```diff
@@ -0,0 +1,82 @@
...
+    secrets:
+      ECR_ROLE_TO_ASSUME:
+        required: true
+    outputs:
+      image_tag:
+        description: "The Docker image tag (commit SHA)"
+        value: ${{ jobs.build.outputs.image_tag }}
+
+jobs:
+  build:
```
sumachalla39, line 19: Looks really good to me, 
Just minor suggestion: Currently the build time is not bad, is it worth using gradle cache to reduce build time in future?

sumachalla39: Agreed with Ben, we can look into this in future when it actually causing slower builds.

sumachalla39 (commented):
LGTM, 
I noticed an unchanged file recommended action on Dockerfile on line 21
Is this something can be fixed here?

sumachalla39 (dismissed):

gradle.properties
```diff
@@ -1,2 +1,11 @@
 version=1.0.0
 group=uk.gov.justice.laa
+
+# Performance optimizations
```
sumachalla39: We cant use build cache in pipeline, cache properties to be removed?

BenMillar-MOJ: Thanks Suma, have removed these

sumachalla39 (approved):

PR #112, opened 2025-12-10
BC-301 Refactor to use the same field for getting the last stored ass…

sumachalla39:
[Link to story](https://dsdmoj.atlassian.net/browse/BC-301)

Show the stored last assessed values on first load of outcome, review pages

- Renamed Status
AmendStatus → AssessmentStatus
Merged NEEDS_ASSESSING and ASSESSMENT status into one.

- Centralised Status Handling
Introduced ClaimStatusHandler to manage status updates. Status updates in ClaimField is removed.
This handler updates the ClaimFields within the provided ClaimDetails object.

- Invocation Points
ClaimStatusHandler is now invoked:
_In AssessmentService after the outcome status is set.
In ClaimReviewController to load status when rendering the page._

- Field Simplification
Removed the amended field from ClaimField.
Using a single assessed attribute consistently across the assessment process.

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/handlers/ClaimStatusHandler.java
```diff
@@ -0,0 +1,164 @@
...
+                claim.getAssessedTotalInclVat(),
+                claim.getAllowedTotalVat(),
+                claim.getAllowedTotalInclVat()
+        );
+
+        List<ClaimField> specificFields = switch (claim) {
+            case CivilClaimDetails civil -> getCivilFields(civil);
+            case CrimeClaimDetails crime -> getCrimeFields(crime);
+            default -> List.of();
+        };
```
jackheslop96: Can we abstract the fields to ClaimDetails / CrimeClaimDetails / CivilClaimDetails so we just need to call, for example, `claim.getFields()`. Feels like there's maybe a bit of overlap between the logic here and the logic in the view models

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ChangeAssessedTotalsController.java
```diff
@@ -40,19 +40,19 @@ public String onPageLoad(
     ) {
         ClaimDetails claim = (ClaimDetails) request.getAttribute(claimId);
 
-        if (claim.getAssessedTotalVat().getStatus() == AmendStatus.DO_NOT_DISPLAY || claim.getAssessedTotalInclVat().getStatus() == AmendStatus.DO_NOT_DISPLAY) {
+        if (claim.getAssessedTotalVat().getStatus() == AssessStatus.DO_NOT_DISPLAY || claim.getAssessedTotalInclVat().getStatus() == AssessStatus.DO_NOT_DISPLAY) {
```
BenMillar-MOJ: What do you think about renaming `AssessStatus` to `ClaimFieldStatus`?

src/main/java/uk/gov/justice/laa/amend/claim/models/ClaimFieldStatus.java
```diff
@@ -0,0 +1,24 @@
...
+ *   <li>{@link #DO_NOT_DISPLAY} - The status should not be shown in the UI.</li>
+ * </ul>
+ */
+
+@Getter
+public enum AssessStatus {
+    ASSESSABLE,
+    NOT_ASSESSABLE,
+    NEEDS_ASSESSING,
+    DO_NOT_DISPLAY
```
BenMillar-MOJ, line 21: Suggestion: What do you think about merging `ASSESSABLE` and `NEEDS_ASSESSING`?

These are fundamentally the same status, but only change how the link is displayed on the table if we use the logic of:

```
if ( AssessmentStatus ==ASSESSABLE) {
  if (value == null) {
    Show add link
  } else {
    Show change link
  }
}
```

We could potentially simplify the logic of setting the `AssessmentStatus` substantially.

BenMillar-MOJ (changes requested):

src/main/java/uk/gov/justice/laa/amend/claim/models/ClaimFieldStatus.java
spennymoj: +1

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ChangeAssessedTotalsController.java
spennymoj: Or maybe something like `ReadOnly` with options of `TRUE` `FALSE` and `HIDDEN` or suchlike

src/main/java/uk/gov/justice/laa/amend/claim/handlers/ClaimStatusHandler.java
sumachalla39: Good Idea, I have extracted the getClaimFields into it's own classes. 

ClaimViewModel changes required bit more attention with lots of testing Scenarios. Will be good to refactor in separate PR

src/main/java/uk/gov/justice/laa/amend/claim/models/ClaimFieldStatus.java
sumachalla39: That sounds good, I will try and merge NEEDS_ASSESSING and ASSESSMENT into one.

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ChangeAssessedTotalsController.java
sumachalla39: Which one is better AssessmentStatus or ClaimFieldStatus?

sumachalla39: Modified to ClaimFieldStatus

BenMillar-MOJ (approved):
Thank you for this!!

PR #143, opened 2026-01-12
BC-265 Added back removed test

sumachalla39:
## What is this PR?

[Link to story](https://dsdmoj.atlassian.net/browse/BC-265)
 Added back the removed unit test to verify bolton fields on summary screen based on requested values. 

## Checklist

### DIFF
--- a/src/test/java/uk/gov/justice/laa/amend/claim/views/ClaimSummaryViewTest.java
+++ b/src/test/java/uk/gov/justice/laa/amend/claim/views/ClaimSummaryViewTest.java
@@ -2,7 +2,11 @@ package uk.gov.justice.laa.amend.claim.views;
 
 import org.jetbrains.annotations.NotNull;
 import org.jsoup.nodes.Document;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
 import org.springframework.context.annotation.Import;
 import org.springframework.security.oauth2.client.OAuth2AuthorizedClient;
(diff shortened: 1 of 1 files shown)


spennymoj:
@jackheslop96 has raised a PR (https://github.com/ministryofjustice/laa-amend-a-claim/pull/144) to a similar effect as this but with a different implementation. I'm partial to Jack's implementation as it uses the ViewTestBase to create a reusable base (which will be useful elsewhere with hidden logic). 

Overall, nothing _wrong_ with this PR, but I'd prefer to go with Jacks if possible.

spennymoj (changes requested):
@jackheslop96 has raised a PR (https://github.com/ministryofjustice/laa-amend-a-claim/pull/144) to a similar effect as this but with a different implementation. I'm partial to Jack's implementation as it uses the ViewTestBase to create a reusable base (which will be useful elsewhere with hidden logic).

Overall, nothing wrong with this PR, but I'd prefer to go with Jacks if possible.

sumachalla39:
> @jackheslop96 has raised a PR (#144) to a similar effect as this but with a different implementation. I'm partial to Jack's implementation as it uses the ViewTestBase to create a reusable base (which will be useful elsewhere with hidden logic).
> 
> Overall, nothing wrong with this PR, but I'd prefer to go with Jacks if possible.

@jackheslop96 @spennymoj It was removed test from the code base and no tests were added to compensate. Hence I have added, it is not new. I am not sure what is wrong in adding Parameterised tests which is essentially doing same. 
For our case we have many scenarios to cover, rather than duplicating with same set of lines in the tests we can make as one test add our scenarios with inputs and expected outputs, this will help us understand better when its broken.. Its easier to fix in one place than multiple places
I am Happy with Jacks PR, as long as it covers all scenarios. Would be nice if its Parameterized tests as we can use same test with different inputs, so that we can avoid fixing multiples tests.
Field verification we can either do label exist or Entire row exist, I happy with any approach. If its used multiple places, ViewTestBase will be nice.

spennymoj:
> > @jackheslop96 has raised a PR (#144) to a similar effect as this but with a different implementation. I'm partial to Jack's implementation as it uses the ViewTestBase to create a reusable base (which will be useful elsewhere with hidden logic).
> > Overall, nothing wrong with this PR, but I'd prefer to go with Jacks if possible.
> 
> @jackheslop96 @spennymoj It was removed test from the code base and no tests were added to compensate. Hence I have added, it is not new. I am not sure what is wrong in adding Parameterised tests which is essentially doing same. For our case we have many scenarios to cover, rather than duplicating with same set of lines in the tests we can make as one test add our scenarios with inputs and expected outputs, this will help us understand better when its broken.. Its easier to fix in one place than multiple places I am Happy with Jacks PR, as long as it covers all scenarios. Would be nice if its Parameterized tests as we can use same test with different inputs, so that we can avoid fixing multiples tests. Field verification we can either do label exist or Entire row exist, I happy with any approach. If its used multiple places, ViewTestBase will be nice.

Yep, I agree that we should have some sort of testing in place around this but as you say, it's essentially doing the same so why have it in twice. Given the other option is something reusable I'm more partial to going down that route especially with testing. To me, doubling the tests doesn't necessarily mean more robust tests, it's just a larger codebase with more changes required when we refactor. As a best practice I'd prefer to done once (but well) than multiple times as it makes refactoring easier. Might be a small issue now but duplicate that across the codebase and we end up with a an array of reduplicated tests which break when we make changes.

PR #144, opened 2026-01-12
BC-265: Additional test coverage that views are being rendered with o…

jackheslop96:
…r without bolt-on fields.

## What is this PR?

## Checklist

### DIFF
--- a/src/test/java/uk/gov/justice/laa/amend/claim/viewmodels/CivilClaimDetailsViewTest.java
+++ b/src/test/java/uk/gov/justice/laa/amend/claim/viewmodels/CivilClaimDetailsViewTest.java
@@ -171,236 +171,321 @@ public class CivilClaimDetailsViewTest extends ClaimDetailsViewTest<CivilClaimDe
 
         @Test
         void rowsRenderedForClaimValuesWhenClaimHasAnAssessment() {
-            ClaimField fixedFee = new ClaimField("1", null, null);
-            ClaimField netProfitCost = new ClaimField("2", null, null);
-            ClaimField netDisbursementAmount = new ClaimField("3", null, null);
-            ClaimField disbursementVatAmount = new ClaimField("4", null, null);
-            ClaimField counselCost = new ClaimField("5", null, null);
-            ClaimField detention = new ClaimField("6", null, null);
-            ClaimField jrFormFilling = new ClaimField("7", null, null);
-            ClaimField adjournedHearing = new ClaimField("8", 100, 100);
-            ClaimField cmrhTelephone = new ClaimField("9", 100, 100);
-            ClaimField cmrhOral = new ClaimField("10", 100, 100);
-            ClaimField hoInterview = new ClaimField("11", 100, 100);
-            ClaimField substantiveHearing = new ClaimField("12", 100, 100);
-            ClaimField vatClaimed = new ClaimField("13", 100, 100);
-            ClaimField totalAmount = new ClaimField("14", 100, 100);
-
             CivilClaimDetails claim = new CivilClaimDetails();
-            claim.setFixedFee(fixedFee);
-            claim.setNetProfitCost(netProfitCost);
-            claim.setNetDisbursementAmount(netDisbursementAmount);
-            claim.setDisbursementVatAmount(disbursementVatAmount);
-            claim.setCounselsCost(counselCost);
-            claim.setDetentionTravelWaitingCosts(detention);
-            claim.setJrFormFillingCost(jrFormFilling);
-            claim.setAdjournedHearing(adjournedHearing);
-            claim.setCmrhTelephone(cmrhTelephone);
-            claim.setCmrhOral(cmrhOral);
-            claim.setHoInterview(hoInterview);
-            claim.setSubstantiveHearing(substantiveHearing);
-            claim.setVatClaimed(vatClaimed);
-            claim.setTotalAmount(totalAmount);
+            claim.setFixedFee(ClaimField.builder().key("ff").build());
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/justice/laa/amend/claim/views/ViewTestBase.java
```diff
@@ -250,6 +250,14 @@ protected void assertPageHasValuesRow(Document doc, String expectedKey, ClaimFie
     Assertions.assertTrue(rowFound);
   }
 
+  protected void assertPageDoesNotHaveValuesRow(Document doc, String expectedKey) {
```
spennymoj: picky but can we rename this? Your naming technically makes sense but pedantically we're checking a _specific_ values row is/isn't present. There _are_ values rows, we're checking one of them isn't there. 

Maybe something like "assertRowIsHidden" as we're asserting the row associated with the expected key is hidden? Whilst the "values row" (of other fields) are present.

src/test/java/uk/gov/justice/laa/amend/claim/views/ClaimSummaryViewTest.java
```diff
@@ -112,10 +112,57 @@ void testCivilClaimPage() throws Exception {
         assertPageHasSummaryListRow(doc, "Case start date", "01 January 2020");
         assertPageHasSummaryListRow(doc, "Case end date", "31 December 2020");
         assertPageHasSummaryListRow(doc, "Date submitted", "15 June 2020 at 09:30:00");
-        assertPageHasValuesRow(doc, "Total", claim.getTotalAmount(), false);
-        assertPageHasValuesRow(doc, "Oral CMRH", claim.getCmrhOral(), false);
-        assertPageHasValuesRow(doc, "Telephone CMRH", claim.getCmrhTelephone(), false);
-        assertPageHasValuesRow(doc, "Counsel costs", claim.getCounselsCost(), false);
+        assertPageHasSummaryListRow(doc, "Total", claim.getTotalAmount(), false);
```
sumachalla39: Just wondering if it would be nice to add scenario based tests as we have many scenarios to test here
1) with boltons on civil, 2) without boltons on civil 3) Partially exist boltons
4) Crime Details with no boltons ( Functionally it should never possible, but just to make sure our code works as expected)

To avoid Duplicating same test code is it worth to add One Test with scenario based, which will make it easy to understand. Where you can supply your ClaimDetails, outcome, expected output to same test method

spennymoj: I wonder if there's a possible option to use the MockClaimsFunctions - i.e. expand it to be able to generate various types of claims (it can currently do civil or crime but could expand to "With bolt ons" and "with some bolt ons" etc then expand the `assertPageHasSummaryListRow` method to something like `assertPageHasCorrectSummaryListItems` and pass in an array of fields to check that the fields are correct. Will likely reduce duplication and make the code reusable too

src/test/java/uk/gov/justice/laa/amend/claim/viewmodels/CrimeClaimDetailsViewTest.java
```diff
@@ -85,104 +85,111 @@ void createMapOfKeyValuePairs() {
...
-
             CrimeClaimDetails claim = new CrimeClaimDetails();
-            claim.setFixedFee(fixedFee);
-            claim.setNetProfitCost(netProfitCost);
-            claim.setNetDisbursementAmount(netDisbursementAmount);
-            claim.setDisbursementVatAmount(disbursementVatAmount);
-            claim.setTravelCosts(travel);
-            claim.setWaitingCosts(waiting);
-            claim.setTotalAmount(totalAmount);
+            claim.setFixedFee(ClaimField.builder().key("ff").build());
```
spennymoj: We have functions in `MockClaimsFunctions` which allow you to create claims. I wonder if you can move this into that class and just create it once and we can use it elsewhere? There should already be an option in there for a full crime/civil claim. Can add a partially filled in claim in there too? 

Small suggestion for reusability :)

src/test/java/uk/gov/justice/laa/amend/claim/viewmodels/CivilClaimDetailsViewTest.java
```diff
@@ -171,236 +171,321 @@ class GetSummaryClaimFieldRowsTests {
...
-            claim.setDetentionTravelWaitingCosts(detention);
-            claim.setJrFormFillingCost(jrFormFilling);
-            claim.setAdjournedHearing(adjournedHearing);
-            claim.setCmrhTelephone(cmrhTelephone);
-            claim.setCmrhOral(cmrhOral);
-            claim.setHoInterview(hoInterview);
-            claim.setSubstantiveHearing(substantiveHearing);
-            claim.setVatClaimed(vatClaimed);
-            claim.setTotalAmount(totalAmount);
+            claim.setFixedFee(ClaimField.builder().key("ff").build());
```
spennymoj: Same applies here as CrimeClaimDetailsViewTest

src/test/java/uk/gov/justice/laa/amend/claim/viewmodels/CrimeClaimDetailsViewTest.java
spennymoj: Also in the `ClaimDetailsViewTest` there's a redeclaring of `CreateClaimField` - which is also already in the `MockClaimsFunctions`. Slightly different implementation tho. Might be worth moving that too

spennymoj (approved):

PR #165, opened 2026-01-22
BC-391: Content security policy

BenMillar-MOJ:
## What is this PR?

[Link to story](https://dsdmoj.atlassian.net/browse/BC-391)

- Adds restricted content security policy to prevent untrusted JS from loading
- Removes references to jQuery, which isn't required an is a potential vulnerability
- Refactors inline JS to `app.js` so all inline JS can be blocked by the CSP

## Checklist


jackheslop96 (approved):
Nice 👍

PR #200, opened 2026-01-30
Feature/bc 400 provider name

sumachalla39:
## What is this PR?

[Link to story](https://dsdmoj.atlassian.net/browse/BC-400)

Upgraded to Java 25 and Gradle 9.2
Deployment of Github runner as separate release and reuse the runners
Modified to deploy ephemeral e2e stack and added cleanup step for feature branches
Added Client to fetch firmName from provider details API

**Provider name shows as here:**
image

**When Provider name not available:**
image

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/ClaimDetailsView.java
```diff
@@ -23,7 +23,7 @@ default Map<String, Object> getSummaryRows() {
         rows.put("clientName", getClientName());
         rows.put("ufn", claim().getUniqueFileNumber());
         addUcnSummaryRow(rows);
-        rows.put("providerName", claim().getProviderName());
+        rows.put("providerName", claim().getProviderName() == null ? new ThymeleafMessage("provider.firmName.notAvailable") : claim().getProviderName());
```
jackheslop96, line 26: Have seen the conversation with Jonathan on the ticket 👍 Can `submissionResponse.getOfficeAccountNumber()` be null? If so then the provider name will ALWAYS be null, so "Currently not available" might be misleading in that scenario.

Also potentially a wider conversation to be had with Jonathan around the use of "Not applicable" on rows with null values if he deems this to not be descriptive enough, but that is separate to this 🙂

sumachalla39: AccountNumber is mandatory field in search, so should submission always contains account number?
Also office_account_nmuber column is NotNull in db. Based on these I think account number always exist unless corrupted data. But we can check with BAs.

jackheslop96: Brill, thank you for confirming 👍

jackheslop96 (dismissed):

jackheslop96 (approved):

PR #203, opened 2026-02-02
feat: add new assessment table

spennymoj:
## What is this PR?

[Link to story](https://dsdmoj.atlassian.net/browse/BC-410)

Added the 'Assessment' table for the "Review and Amend" Screen as per the new designs.
Removed the back line on this screen only
Updated unit tests and E2E tests accordingly.

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/models/Claim.java
```diff
@@ -42,4 +43,8 @@ public class Claim implements Serializable {
     public BaseClaimView<? extends Claim> toViewModel() {
         return new ClaimView(this);
     }
+
+    public String reviewAssessmentChangeUrl(String submissionId, String claimId) {
+        return String.format("/submissions/%s/claims/%s/assessment-outcome",  submissionId, claimId);
+    }
```
jackheslop96: With this being UI logic can it live in the view model instead?

src/main/resources/templates/review-and-amend.html
```diff
@@ -5,7 +5,7 @@
         title=#{claimReview.title},
         mainContent=~{::#main-content},
         pageCategory=${null},
-        showBackLink=${true},
+        showBackLink=${false},
```
jackheslop96: Now that we've set this to false can we remove the `backLinkUrl` stuff?

src/main/resources/templates/review-and-amend.html
```diff
@@ -28,6 +28,62 @@ <h1 class="govuk-heading-xl">
         <span th:text="#{${(claim.assessmentOutcome != null) ? claim.assessmentOutcome.messageKey : 'claimValuesTable.noData'}}"></span>
```
BenMillar-MOJ: I believe this should now be removed, as we are showing the assessment outcome in the table.

src/main/resources/templates/review-and-amend.html
```diff
@@ -28,6 +28,62 @@ <h1 class="govuk-heading-xl">
...
+                                class="govuk-link govuk-link--no-visited-state">
+                            <th:block th:text="#{service.change}"></th:block>
+                            <span class="govuk-visually-hidden" th:text="| #{claimSummary.rows.assessmentOutcome}|"></span>
+                        </a>
+                    </td>
+                </tr>
+
+                <tr class="govuk-table__row">
+
+                    <td class="govuk-table__cell govuk-!-font-weight-bold" th:text="#{claimSummary.rows.vatLiability}"></td>
```
BenMillar-MOJ, line 59: This should show the assessed value for VAT liability, rather than the original value.

src/main/resources/templates/review-and-amend.html
```diff
@@ -28,6 +28,62 @@ <h1 class="govuk-heading-xl">
...
+
+
+                    <td class="govuk-table__cell" colspan="3"
+                        th:text="${@ThymeleafUtils.getFormattedValue(claim.vatClaimed.getSubmitted()).resolve(#messages)}"/>
+
+                    <td class="govuk-table__cell">
+                        <a
+                           th:href="${claim.reviewAssessmentChangeUrl(submissionId, claimId)}"
```
BenMillar-MOJ: This should include an anchor tag to the VAT liability question.

src/main/resources/templates/review-and-amend.html
```diff
@@ -28,6 +28,62 @@ <h1 class="govuk-heading-xl">
...
+
+                <tr class="govuk-table__row">
+
+                    <td class="govuk-table__cell govuk-!-font-weight-bold" th:text="#{claimSummary.rows.assessmentOutcome}"></td>
+
+                    <td class="govuk-table__cell" colspan="3" th:text="#{${(claim.assessmentOutcome != null) ? claim.assessmentOutcome.messageKey : 'claimValuesTable.noData'}}"/>
+
+                    <td class="govuk-table__cell">
+                        <a
+                                th:href="${claim.reviewAssessmentChangeUrl(submissionId, claimId)}"
```
BenMillar-MOJ: This should include an anchor tag to the assessment outcome question.

src/main/resources/templates/review-and-amend.html
```diff
@@ -28,6 +28,62 @@ <h1 class="govuk-heading-xl">
...
+
+                    <td class="govuk-table__cell" colspan="3"
+                        th:text="${@ThymeleafUtils.getFormattedValue(claim.vatClaimed.getSubmitted()).resolve(#messages)}"/>
+
+                    <td class="govuk-table__cell">
+                        <a
+                           th:href="${claim.reviewAssessmentChangeUrl(submissionId, claimId)}"
+                           th:data-testid="claim-field-vat-liability"
+                           class="govuk-link govuk-link--no-visited-state">
+                            <th:block th:text="#{service.change}"></th:block>
```
BenMillar-MOJ, line 69: Why is this in another block rather than adding the text to the link tag?

BenMillar-MOJ (changes requested):
Thanks for this, have left a few small comments

src/main/resources/templates/review-and-amend.html
BenMillar-MOJ: Great point, these are the changes which can be reverted: https://github.com/ministryofjustice/laa-amend-a-claim/pull/163/changes

src/main/resources/static/css/styles.css
```diff
@@ -25,14 +25,14 @@
 }
 
 .govuk-table__row-item{
-    min-width: 14.75rem;
+    width: 18rem;
```
BenMillar-MOJ, line 29: Could you explain in a comment why we have this and why we are using these values please

src/main/resources/templates/review-and-amend.html
spennymoj: This was something @jackheslop96 and I discussed - because of Thymeleaf being Thymeleaf if you use only a single `th:text` nested elements aren't rendered - so to render the "visually hidden text" (accessibility) they need to be put in blocks.

src/main/resources/static/css/styles.css
spennymoj: On the ticket? Sure.

For reference here though - the `min-widths` were brought in for the alignment. It was a relatively straightforward fix for when the tables all had similar column numbers. Introducing this new table highlighted an easier way to do it. With the 'width' property instead the tables are resizeable to much smaller still retaining their alignments. The sizes used were picked on this logic:

1. Actions - slightly larger than 'Change' - which will always be the 'largest value'
2. Values - larger than 'Not Applicable' - giving room for large numbers also.
3. Item - using the remaining size appropriately.

src/main/resources/templates/review-and-amend.html
spennymoj: good spot!

src/main/resources/templates/review-and-amend.html
jackheslop96: @spennymoj further to your changes we can also remove the `th:if="${showBackLink and backLinkUrl != null}"` case in `layout.html`

spennymoj: @jackheslop96 - fixed that too

src/main/resources/templates/layout.html
```diff
@@ -21,11 +21,6 @@
                 </p>
             </div>
 
-            <a th:if="${showBackLink and backLinkUrl != null}"
-               th:href="${backLinkUrl}"
-               class="govuk-back-link"
-               th:text="#{service.back}">
-            </a>
             <a th:if="${showBackLink and backLinkUrl == null}"
```
jackheslop96: This can now just be `<a th:if="${showBackLink}"`

src/main/resources/static/css/styles.css
BenMillar-MOJ: Brilliant, thank you very much for this. Would you be able to add the reasons for the sizing to the css file please, as that will help massively when it comes to understand why these decisions were made in the future.

BenMillar-MOJ (approved):
LGTM, thanks for this!

PR #204, opened 2026-02-02
BC-250: a11y checks

jackheslop96:
## Axe reporting

[Link to story](https://dsdmoj.atlassian.net/browse/BC-250)

- Using deque axe-core: https://docs.deque.com/devtools-for-web/4/en/java-test-playwright
- Abstracted `waitForPage()` to perform axe check on a page load
- Before collecting results we check to see if the page has previously been analysed by saving the JSON as the heading text
- After tests have finished, a gradle task uses the `axe-html-reporter` (https://github.com/lpelypenko/axe-html-reporter) to convert each JSON file into HTML
- These files are then saved as artifacts and can be downloaded through the github workflow

Screenshot 2026-02-03 at 08 37 31

Screenshot 2026-02-03 at 08 44 53

https://github.com/ministryofjustice/laa-amend-a-claim/actions/runs/21623033614/artifacts/5354450357

## Checklist


.github/workflows/e2e-tests.yml
```diff
@@ -168,6 +168,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6.0.2
 
+      - name: Set up Node.js
```
sumachalla39: The runner image is using an older Node.js version. Would it make sense to update the Dockerfile used in the e2e tests so we don’t need to install Node.js during the workflow?

sumachalla39 (commented):
Looks good to me, a minor comment

.github/workflows/e2e-tests.yml
jackheslop96: Good spot @sumachalla39 yes will make this change

sumachalla39 (approved):

PR #206, opened 2026-02-03
BC-408

jackheslop96:
## BC-408

[Link to story](https://dsdmoj.atlassian.net/browse/BC-408)

- Ensuring `page` and `sort` are cached in search url
- Validating form on GET as defensive measure against putting query params in the URL directly
- Returning bad request for:
  - Invalid sort field (i.e. not one of uniqueFileNumber, caseReferenceNumber or scheduleReference)
  - Invalid sort direction (i.e. not one of asc or desc)
  - Invalid query parameter
- Small bug fix to show "1 search result" instead of "1 search results"

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/models/Sort.java
```diff
@@ -30,13 +32,13 @@ public static Sort defaults() {
 
     public Sort(String str) {
         if (str != null) {
-            Pattern pattern = Pattern.compile("^(\\w+),(\\w+)$");
+            Pattern pattern = Pattern.compile("^(uniqueFileNumber|caseReferenceNumber|scheduleReference),(asc|desc)$");
```
jackheslop96: Tidy this up

src/test/java/uk/gov/justice/laa/amend/claim/views/IndexViewTest.java
```diff
@@ -74,30 +79,57 @@ void testPage() throws Exception {
...
+        assertPageHasH2(doc, "20 search results");
 
         assertPageHasTable(doc);
 
         assertPageHasPagination(doc);
+
+        Elements headers = getTableHeaders(doc);
+
+        assertTableHeaderIsNotSortable(headers.get(0), "Claim");
+        assertTableHeaderIsSortable(headers.get(1), "ascending", "UFN", "/?page=2&sort=uniqueFileNumber,desc");
```
jackheslop96: Change this functionality back so they go to page 1

src/main/resources/messages.properties
```diff
@@ -51,6 +51,7 @@ index.caseReferenceNumber.error.invalid=The CRN must only include letters a to z
 
 index.results.none=There are no results that match the search criteria
 index.results.retry=Check you've entered the correct details
+index.results.searchResult=1 search result
 index.results.searchResults={0} search results
```
BenMillar-MOJ, line 55: I think this can be refactored like this:

`index.results.searchResults={0,choice,1#1 search result|1<{0} search results}`

Then you shouldn't need to handle the logic in Thymeleaf.

BenMillar-MOJ (changes requested):

src/main/resources/messages.properties
jackheslop96: What is that syntax 🙈 I'm actually tempted to leave as is if that's ok? I think someone picking up thymeleaf for the first time would have better luck understanding the ternary in the th:text

jackheslop96: @BenMillar-MOJ we chatted about this in tech standup. Going to leave as is if you're ok with that? Couple of reasons:

1. Readability
2. Ease of translation if the need ever arose in the future for Welsh translation

BenMillar-MOJ: Yeah, that's fair! Sounds good!

BenMillar-MOJ (approved):
LGTM, thank you for this!

PR #230, opened 2026-02-12
BC-322: Created a not found error screen and a generic error screen.

jackheslop96:
## BC-322

[Link to story](https://dsdmoj.atlassian.net/browse/BC-322)

Created a not found error screen and a generic error screen.

screencapture-localhost-8080-2026-02-12-17_32_43

screencapture-localhost-8080-foo-2026-02-12-17_27_47

## Checklist

### DIFF
--- a/src/main/java/uk/gov/justice/laa/amend/claim/controllers/MaintenancePageController.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/controllers/MaintenancePageController.java
@@ -1,20 +1,26 @@
 package uk.gov.justice.laa.amend.claim.controllers;
 
 import lombok.AllArgsConstructor;
+import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.server.ResponseStatusException;
 import uk.gov.justice.laa.amend.claim.service.MaintenanceService;
 
 @Controller
 @AllArgsConstructor
 public class MaintenancePageController {
+
     private final MaintenanceService maintenanceService;
 
     @GetMapping("/maintenance")
     public String onPageLoad(Model model) {
-        model.addAttribute("maintenanceMessage", maintenanceService.getMessage());
-        model.addAttribute("maintenanceTitle", maintenanceService.getTitle());
-        return "maintenance";
+        if (maintenanceService.maintenanceEnabled()) {
+            model.addAttribute("maintenanceMessage", maintenanceService.getMessage());
+            model.addAttribute("maintenanceTitle", maintenanceService.getTitle());
+            return "maintenance";
+        }
+        throw new ResponseStatusException(HttpStatus.NOT_FOUND);
     }
 }
--- a/src/main/java/uk/gov/justice/laa/amend/claim/controllers/MaintenancePageController.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/controllers/MaintenancePageController.java
@@ -1,0 +1,20 @@
+package uk.gov.justice.laa.amend.claim.controllers;
+
+import lombok.AllArgsConstructor;
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.GetMapping;
+import uk.gov.justice.laa.amend.claim.service.MaintenanceService;
+
+@Controller
+@AllArgsConstructor
+public class MaintenancePageController {
+    private final MaintenanceService maintenanceService;
+
+    @GetMapping("/maintenance")
+    public String onPageLoad(Model model) {
+        model.addAttribute("maintenanceMessage", maintenanceService.getMessage());
+        model.addAttribute("maintenanceTitle", maintenanceService.getTitle());
+        return "maintenance";
+    }
+}
--- a/src/main/java/uk/gov/justice/laa/amend/claim/exceptions/ErrorPageController.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/exceptions/ErrorPageController.java
@@ -2,21 +2,40 @@ package uk.gov.justice.laa.amend.claim.exceptions;
 
 import jakarta.servlet.RequestDispatcher;
 import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import java.util.Optional;
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.boot.webmvc.error.ErrorController;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
 import org.springframework.web.bind.annotation.RequestMapping;
+import uk.gov.justice.laa.amend.claim.factories.ReferenceNumberFactory;
+import uk.gov.justice.laa.amend.claim.models.ReferenceNumber;
 
 @Controller
+@AllArgsConstructor
+@Slf4j
 public class ErrorPageController implements ErrorController {
 
+    private final ReferenceNumberFactory referenceNumberFactory;
+
     @RequestMapping("/error")
-    public String handleError(HttpServletRequest request, Model model) {
-        Object status = request.getAttribute(RequestDispatcher.ERROR_STATUS_CODE);
-        if (status != null) {
-            int statusCode = Integer.parseInt(status.toString());
-            model.addAttribute("statusCode", statusCode);
+    public String handleError(HttpServletRequest request, HttpServletResponse response, Model model) {
+        int status = Optional.ofNullable(request.getAttribute(RequestDispatcher.ERROR_STATUS_CODE))
+                .map(x -> Integer.parseInt(x.toString()))
+                .orElse(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+        response.setStatus(status);
+        if (status == HttpServletResponse.SC_NOT_FOUND) {
+            return "not-found";
         }
+        ReferenceNumber referenceNumber = referenceNumberFactory.create();
+        model.addAttribute("referenceNumber", referenceNumber);
+        log.error(
+                "Something went wrong. Reference: {}. Status: {}. Session ID: {}",
+                referenceNumber,
+                status,
+                request.getSession().getId());
         return "error";
     }
 }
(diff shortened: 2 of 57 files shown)


src/main/resources/templates/error.html
```diff
@@ -1,23 +1,24 @@
...
+    <body>
+        <main class="govuk-main-wrapper" id="main-content">
+            <div class="govuk-grid-row">
+                <div class="govuk-grid-column-two-thirds">
+                    <h1 class="govuk-heading-l" th:text="#{error.heading}"></h1>
+                    <p class="govuk-body" th:text="#{error.p1}"></p>
+                    <p class="govuk-body" th:text="#{error.p2}"></p>
+                    <p class="govuk-body">
+                        <th:block th:text="#{error.p3.1}"></th:block>
+                        <a class="govuk-link" id="email" th:href="'mailto:' + ${@environment.getProperty('app.email')}" th:text="#{error.p3.link}"></a>
```
mboyd-moj: mailto link should include the reference

src/main/java/uk/gov/justice/laa/amend/claim/exceptions/ErrorPageController.java
```diff
@@ -2,21 +2,40 @@
...
 
+    private final ReferenceNumberFactory referenceNumberFactory;
+
     @RequestMapping("/error")
-    public String handleError(HttpServletRequest request, Model model) {
-        Object status = request.getAttribute(RequestDispatcher.ERROR_STATUS_CODE);
-        if (status != null) {
-            int statusCode = Integer.parseInt(status.toString());
-            model.addAttribute("statusCode", statusCode);
+    public String handleError(HttpServletRequest request, HttpServletResponse response, Model model) {
```
mboyd-moj, line 23: Did you look at @ExceptionHandler? My hope was that we could just have a separate method to handle 404 exceptions so we don't need to do this digging into the innards of the request

src/main/java/uk/gov/justice/laa/amend/claim/factories/ReferenceNumberFactory.java
```diff
@@ -0,0 +1,14 @@
...
+
+import org.apache.commons.lang3.RandomStringUtils;
+import org.springframework.stereotype.Component;
+import uk.gov.justice.laa.amend.claim.models.ReferenceNumber;
+
+@Component
+public class ReferenceNumberFactory {
+
+    public ReferenceNumber create() {
+        String referenceNumber = RandomStringUtils.secure().next(6, true, true).toUpperCase();
```
mboyd-moj: can we just use a uuid? They'll just be copy-pasting it so not worried about making it short

src/main/java/uk/gov/justice/laa/amend/claim/models/ReferenceNumber.java
```diff
@@ -0,0 +1,3 @@
+package uk.gov.justice.laa.amend.claim.models;
+
+public record ReferenceNumber(String value) {}
```
mboyd-moj: Feels like we can just remove this class

src/main/resources/templates/error.html
jackheslop96: Sorry yes I missed that in the ACs. Seems to directly contradict the content in the comments though 🤔 

> Contact the Amend a Bulk Claim digital team (Insert mailto link) quoting reference: [insert error reference] if you have any questions

mboyd-moj: I think it doesn't hurt to have both? When they click the link it should prepopulate just to save them some time, and if they somehow delete or fail to copy that across then they've got the instruction to copy it in

src/main/java/uk/gov/justice/laa/amend/claim/factories/ReferenceNumberFactory.java
jackheslop96: Happy to have a conversation with Jonathan about this and how a UUID would look on the screen (does it draw attention away from the rest of the content, is it a screen-reader-friendly approach etc.)

src/main/resources/templates/error.html
jackheslop96: I can stick the reference number in the subject if we think that would suffice?

mboyd-moj: Yeah that's what I was thinking, just having it prepopulate the subject sounds good

src/main/java/uk/gov/justice/laa/amend/claim/factories/ReferenceNumberFactory.java
mboyd-moj: You could always have it underneath if it helps. I feel like often debuggy stuff on error pages is underneath and sometimes even slighly greyed to keep attention on the main text

mboyd-moj (changes requested):

src/main/java/uk/gov/justice/laa/amend/claim/exceptions/ErrorPageController.java
jackheslop96: @mboyd-moj I've looked into exception handler. I think ultimately we'd end up with something similar to what we already have. While you can define a controller advice like

```
@ControllerAdvice
class GlobalControllerExceptionHandler {

    @ResponseStatus(HttpStatus.NOT_FOUND)
    @ExceptionHandler(Exception.class)
    public ModelAndView handleNotFound() {
        ...
    }
}
```

You can't then also define a separate "catch all" like

```
@ControllerAdvice
class GlobalControllerExceptionHandler {

    @ResponseStatus(HttpStatus.NOT_FOUND)
    @ExceptionHandler(Exception.class)
    public ModelAndView handleNotFound() {
        ...
    }

    @ExceptionHandler(Exception.class)
    public ModelAndView handleError() {
        ...
    }
}
```

As the two exception handlers would get in each other's way. So you'd need a single handler like:

```
@ControllerAdvice
class GlobalControllerExceptionHandler {

    @ExceptionHandler(Exception.class)
    public ModelAndView handleError() {
        // if status is 404 display not found page otherwise display error page
        ...
    }
}
```

jackheslop96: Happy to pair on it though if you had something else in mind 🙂

mboyd-moj: I believe you can use your catchall by making using of the @Order annotation
see https://codemia.io/knowledge-hub/path/setting_precedence_of_multiple_controlleradvice_exceptionhandlers

mboyd-moj: It reads as if that annotation should be applicable to methods too so hopefully wouldn't need multiple controller advice

mboyd-moj: Looked into this offline and came to the conclusion ControllerAdvice is not appropriate due to some errors not reaching it. Sorry for the wild goose chase, your solution looks good :)

src/main/java/uk/gov/justice/laa/amend/claim/exceptions/ErrorPageController.java
```diff
@@ -2,21 +2,39 @@
...
+        int status = Optional.ofNullable(request.getAttribute(RequestDispatcher.ERROR_STATUS_CODE))
+                .map(x -> Integer.parseInt(x.toString()))
```
mboyd-moj, line 25: Are we happy to expose all error codes back to the user? Might be better to have a set that we explicitly handle (i.e. just 404 for now) and fallback to 500 if not handled. Just thinking if we risk exposing unintended information

src/main/java/uk/gov/justice/laa/amend/claim/controllers/MaintenancePageController.java
```diff
@@ -1,20 +1,26 @@
...
     public String onPageLoad(Model model) {
-        model.addAttribute("maintenanceMessage", maintenanceService.getMessage());
-        model.addAttribute("maintenanceTitle", maintenanceService.getTitle());
-        return "maintenance";
+        if (maintenanceService.maintenanceEnabled()) {
+            model.addAttribute("maintenanceMessage", maintenanceService.getMessage());
+            model.addAttribute("maintenanceTitle", maintenanceService.getTitle());
+            return "maintenance";
+        }
+        throw new ResponseStatusException(HttpStatus.NOT_FOUND);
```
mboyd-moj, line 24: nice find :)

src/main/resources/messages.properties
```diff
@@ -300,4 +300,19 @@ signOut.title=You are now signed out of your account
 signOut.message=Sign in to continue using the service.
 sessionExpired.heading=For your security, we signed you out
 signOut.heading=You are now signed out of your account
-sessionExpired.message=This is because you were inactive for 15 minutes.
\ No newline at end of file
+sessionExpired.message=This is because you were inactive for 15 minutes.
+
+notFound.title=Page not found
+notFound.heading=Page not found
+notFound.p1=If you entered a web address please check it was correct.
```
mboyd-moj, line 307: nit: missing a comma after address

src/test/java/uk/gov/justice/laa/amend/claim/controllers/MaintenancePageControllerTest.java
```diff
@@ -0,0 +1,61 @@
...
+                .andExpect(model().attribute("maintenanceTitle", title));
+    }
+
+    @Test
+    public void testOnPageLoadReturnsNotFoundWhenMaintenanceDisabled() throws Exception {
+        when(maintenanceService.maintenanceEnabled()).thenReturn(false);
+
+        mockMvc.perform(get("/maintenance")).andExpect(status().isNotFound());
+
+        verify(maintenanceService, never()).getMessage();
```
mboyd-moj, line 58: `verifyNoMoreInteractions(maintenanceService)` is probably simpler

src/test/java/uk/gov/justice/laa/amend/claim/factories/ReferenceNumberFactoryTest.java
```diff
@@ -0,0 +1,35 @@
+package uk.gov.justice.laa.amend.claim.factories;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.RepeatedTest;
+
+public class ReferenceNumberFactoryTest {
+
+    @RepeatedTest(10)
```
mboyd-moj, line 8: RepeatedTest is a bit excessive here as having a non-constant length seems very low risk

mboyd-moj (approved):

PR #245, opened 2026-02-18
BC-461: Added zap action

jackheslop96:
## BC-461

[Link to story](https://dsdmoj.atlassian.net/browse/BC-461)

- Introduced zap worflow
- Runs daily at midnight
- Runs the application on localhost, spinning up redis and wiremock in docker
- Sends message to Slack on failure
- Report (`zap_scan`) and application logs (`zap_output.log`) are uploaded as artifacts
- Target URL set as `http://localhost:8080/?providerAccountNumber=0P322F` so that ZAP can see the view links that allow it to discover subsequent pages/URLs
- In the context file (`zap.context`) we tell ZAP that everything on `http://localhost:8080/.*` is in scope
- `rules.tsv` file allows us to ignore certain alerts

## Checklist


ahmadsjustice (dismissed):

.github/workflows/zap.yml
```diff
@@ -0,0 +1,81 @@
...
+        uses: zaproxy/action-full-scan@v0.13.0 # https://github.com/zaproxy/action-full-scan
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          target: 'http://localhost:8080/?providerAccountNumber=0P322F' # Exposes 'View' links for ZAP spider to discover
+          rules_file_name: '.zap/rules.tsv'
+          cmd_options: '-z "-config spider.maxDepth=10" -n zap.context' # https://www.zaproxy.org/docs/docker/full-scan/#usage
+          allow_issue_writing: false
+          fail_action: true
+      - name: Upload logs
+        uses: actions/upload-artifact@v6.0.0
```
ahmadsjustice, line 80: do we need to add if: always() 
Otherwise The upload step won't run if the ZAP scan step fails ?

.github/workflows/zap.yml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          name: zap_output.log
+          path: zap_output.log
+          retention-days: 1
```
ahmadsjustice: ZAP also generates a html report ? could we add an upload for the report too please.

.github/workflows/zap.yml
jackheslop96: Good call, yes I will do that

.github/workflows/zap.yml
jackheslop96: action-full-scan does this for us. It ends up in the artifacts along with the logs.

mboyd-moj (approved):

PR #252, opened 2026-02-19
BC-210: Use V2 features of get claims endpoint

mboyd-moj:
## What is this PR?

https://dsdmoj.atlassian.net/browse/BC-210

Added filters for Area of Law and Escape Case
- Did not include No as an Escape Case option because this can be either null or false in data claims

Made more columns sortable
- Did not include sorting of Escape Case column because null values in here mean the backend sorting includes null/true/false

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/models/Sort.java
```diff
@@ -26,7 +26,7 @@ public static Sort defaults() {
 
     public Sort(String str) {
         if (str != null) {
-            Pattern pattern = Pattern.compile("^(\\w+),(\\w+)$");
+            Pattern pattern = Pattern.compile("^([\\w.]+),(\\w+)$");
```
jackheslop96: What's the reason behind this change?

jackheslop96: Ah I see because the key can be e.g. client.clientSurname

Should the . be escaped? Is it currently meaning all characters?

jackheslop96:
Can the wiremocks be updated to reflect the api version change for certain endpoints? Otherwise really good 👍

src/main/java/uk/gov/justice/laa/amend/claim/models/Sort.java
mboyd-moj: I think it's OK because it's inside square brackets

build.gradle
```diff
@@ -44,7 +44,8 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-security-oauth2-resource-server'
 
     // Data claims API
-    implementation 'uk.gov.justice.laa.datastewardship.payments:data-claims-api-models:d694928'
+    // TODO: Revert to main
```
mboyd-moj: TODO

e2e.sh
```diff
@@ -28,7 +28,8 @@ export SILAS_AUTH_ENABLED=false
 
 echo "[INFO] Starting API application..."
 pushd ../laa-data-claims-api >/dev/null
-git checkout main
+# TODO: Revert to main
```
mboyd-moj: TODO

src/main/java/uk/gov/justice/laa/amend/claim/mappers/ClaimMapper.java
```diff
@@ -134,8 +129,7 @@ default YearMonth mapSubmissionPeriod(ClaimResponse claimResponse) {
     }
 
     private void enrichWithSubmission(ClaimDetails claim, SubmissionResponse submissionResponse) {
-        if (submissionResponse != null && submissionResponse.getAreaOfLaw() != null) {
-            claim.setAreaOfLaw(submissionResponse.getAreaOfLaw().getValue());
+        if (submissionResponse != null) {
```
sumachalla39: Just checking do we still need to use SubmissionResponse? providerAccountNumber, submitteddate not available in ClaimResposeV2?

src/main/java/uk/gov/justice/laa/amend/claim/models/SortField.java
```diff
@@ -7,7 +7,10 @@
 public enum SortField {
     UNIQUE_FILE_NUMBER("uniqueFileNumber"),
     CASE_REFERENCE_NUMBER("caseReferenceNumber"),
-    SCHEDULE_REFERENCE("scheduleReference");
+    CLIENT_SURNAME("client.clientSurname"),
+    SUBMISSION_PERIOD("submission.submissionPeriod"),
+    SCHEDULE_REFERENCE("scheduleReference"),
+    CATEGORY_OF_LAW("calculatedFeeDetail.categoryOfLaw");
```
sumachalla39: We won't be using categoryOfLaw field on Search page? Categoryoflaw should be renamed to areaOfLaw? My understanding is areaOfLaw comes from submission response in V1 submission api in the previous version, now it should be from ClaimResponseV2 of Claims api?

mboyd-moj: Chatted this through with Ben/Jonathon/Chris and we decided it was best to leave the column as Category of Law in the search results for now as this was what was signed off in UAT. So we're just adding the filters and making columns sortable where possible

src/main/java/uk/gov/justice/laa/amend/claim/mappers/ClaimMapper.java
mboyd-moj: It was just submittedDate left, but Jose is going to add it so we can remove this enrich logic completely 😍

mboyd-moj: Removed the call to submission API 🥳

jackheslop96 (approved):

PR #267, opened 2026-02-27
BC-487: Add page for void confirmation

mboyd-moj:
## What is this PR?

https://dsdmoj.atlassian.net/browse/BC-487

BC-487: Add page for void confirmation. Page can only be navigated to directly via URL for now until previous page is implemented

## Checklist

### DIFF
--- a/src/main/java/uk/gov/justice/laa/amend/claim/controllers/VoidConfirmationController.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/controllers/VoidConfirmationController.java
@@ -14,6 +14,7 @@ import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import uk.gov.justice.laa.amend.claim.models.ClaimDetails;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.ClaimStatus;
 
 @Controller
 @RequestMapping("/submissions/{submissionId}/claims/{claimId}/void")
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/justice/laa/amend/claim/controllers/VoidConfirmationController.java
```diff
@@ -0,0 +1,93 @@
...
+            HttpSession session,
+            Model model,
+            @PathVariable UUID submissionId,
+            @PathVariable UUID claimId,
+            HttpServletResponse response)
+            throws IOException {
+        if (!isVoidingEnabled) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND);
+            return null;
+        }
```
sumachalla39, line 58: Can feature verification moved to seperate interceptor to handle void uri requests to avoid duplication on get or post methods?

src/main/java/uk/gov/justice/laa/amend/claim/controllers/VoidConfirmationController.java
```diff
@@ -0,0 +1,93 @@
...
+        return "void-confirmation";
+    }
+
+    private ClaimDetails getValidClaim(HttpSession session, UUID claimId) {
+        var claim = (ClaimDetails) session.getAttribute(claimId.toString());
+        if (claim.getStatus() != ClaimStatus.VALID) {
+            throw new IllegalStateException("Claim status is not VALID");
+        }
+        return claim;
+    }
```
sumachalla39, line 92: getValidClaim also can be moved into interceptor of void url

mboyd-moj: This one is just a sanity check in case they came here directly. Not sure there is value in adding a separate interceptor for the sake of a couple of lines of code

src/main/java/uk/gov/justice/laa/amend/claim/controllers/VoidConfirmationController.java
mboyd-moj: I prefer having these inline as they're just simple checks and will be removed when the feature flags are removed

sumachalla39 (approved):

PR #278, opened 2026-03-03
Feature/bc 464 void claim details

sumachalla39:
## What is this PR?

[Link to story](https://dsdmoj.atlassian.net/browse/BC-464)

Updated Claim Details Page to 
- show banner for Voided claims
- Hide Add/Update assessment buttons
- Hide Void claim button.

Added mock responses to show void response.
e2e tests - Updated with claims-api image of new api changes.

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/service/AssessmentService.java
```diff
@@ -107,15 +111,28 @@ public CreateAssessment201Response submitAssessment(ClaimDetails claim, String u
     }
 
     public ClaimDetails getLatestAssessmentByClaim(ClaimDetails claimDetails) {
+        // get latest 2 assessments for VOID Claim
+        boolean isVoid = claimDetails.getStatus() == ClaimStatus.VOID;
+        var size = isVoid ? 2 : 1;
```
mboyd-moj: Is it safe for us to rely on it always being the second to last assessment that we care about in a void case? To me it feels like safer approaches would be to either:
- Add the option to filter by assessment type to the data-claims-api
- Get all assessments and filter on our side for the latest one with the escape claim status

I'd prefer the former. But if we can't get that done then the latter option would be preferable to me vs what we have here as it's more robust

build.gradle
```diff
@@ -45,7 +45,7 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-jackson'
 
     // API models
-    implementation('uk.gov.justice.laa.datastewardship.payments:data-claims-api-models:d694928')
+    implementation('uk.gov.justice.laa.datastewardship.payments:data-claims-api-models:c8d6279-SNAPSHOT')
```
mboyd-moj: Just calling out that we shouldn't merge until this has been replaced with the up to date models from main

src/main/resources/templates/partials/information-alert.html
```diff
@@ -1,12 +1,25 @@
-<th:block th:fragment="informationAlert(prefix,content)">
-    <div role="region" class="moj-alert moj-alert--information" th:with="key=${prefix} + '.alert.ariaLabel'" th:aria-label="#{${key}}" data-module="moj-alert">
+<th:block th:fragment="informationAlert(prefix, content, type)">
+    <div data-module="moj-alert" role="region"
+         th:aria-label="#{${key}}"
+         th:class="${type == 'information'} ? 'moj-alert moj-alert--information' : 'moj-alert moj-alert--error'"
```
jackheslop96: We already have the `error-alert` component. Could we handle this logic solely in claim-summary? e.g. if voided display the error component, if assessed display the information component

src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/ClaimDetailsView.java
```diff
@@ -112,16 +114,30 @@ default AssessmentInfo lastAssessment() {
     }
 
     default ThymeleafMessage lastEditedBy(MicrosoftApiUser user) {
+        var voidMessage = new ThymeleafMessage("claimSummary.void.message");
```
jackheslop96, line 117: Unit tests?

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
```diff
@@ -33,7 +35,7 @@ public class ClaimSummaryController {
     public String onPageLoad(
             HttpSession session, Model model, @PathVariable UUID submissionId, @PathVariable UUID claimId) {
         ClaimDetails claim = claimService.getClaimDetails(submissionId, claimId);
-        if (claim.getStatus() != ClaimStatus.VALID) {
+        if (!EnumSet.of(VALID, VOID).contains(claim.getStatus())) {
```
jackheslop96, line 38: Very minor but could stick something like this in `Claim` for readability and reuse

```
public boolean isViewable() {
        return EnumSet.of(VALID, VOID).contains(status);
    }
```

src/main/java/uk/gov/justice/laa/amend/claim/models/AssessmentTypeEnum.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.laa.amend.claim.models;
+
+import lombok.Getter;
+
+@Getter
+public enum AssessmentTypeEnum {
+    ESCAPE_CASE_ASSESSMENT("ESCAPE_CASE_ASSESSMENT"),
+    VOID("VOID");
+
+    private final String value;
```
jackheslop96: Is `value` serving any practical benefit at the moment or could this just be
```
public enum AssessmentTypeEnum {
    ESCAPE_CASE_ASSESSMENT,
    VOID
}
```
?

.github/workflows/e2e-tests.yml
```diff
@@ -20,7 +20,7 @@ on:
       claims_api_image_tag:
         required: false
         type: string
-        default: 'd694928c4891a615dd2658abe08d7be41c1db226'
+        default: 'bb6f8c5462220be10d20f6001a5079ffde4ade2f'
```
mboyd-moj: Just calling out that we shouldn't merge until this has been replaced with the up to date image from main

src/main/java/uk/gov/justice/laa/amend/claim/service/AssessmentService.java
jackheslop96: Just keen to understand the logic here - I can't see anything in the ACs around retrieving the latest two assessments. Do you mind explaining?

src/main/resources/templates/claim-summary.html
```diff
@@ -123,6 +131,15 @@ <h2 class="govuk-summary-card__title" th:text="#{claimSummary.allowedTotalsTable
                                     </button>
                                 </th:block>
 
+                                <th:block sec:authorize="hasRole('ROLE_CLAIM_AMENDMENTS_CASEWORKER')" th:if="${!claim.isVoidClaim}">
```
jackheslop96, line 130: Apologies if I've missed this somewhere but would be cool to see this tested with @WithMockUser(roles = "ROLE_CLAIM_AMENDMENTS_CASEWORKER") if that's possible

src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/ClaimDetailsView.java
```diff
@@ -141,4 +157,8 @@ private Object getProviderName() {
         }
         return claim().getProviderName();
     }
+
+    default boolean isVoidClaim() {
+        return ClaimStatus.VOID == claim().getStatus();
+    }
```
jackheslop96, line 163: If you stick this method in the Claim model itself then you can reuse it in `getLatestAssessmentByClaim`

src/main/resources/templates/claim-summary.html
mboyd-moj: Could we defer this to the permissions ticket? That may change how permissions are done. I think my ticket will move this logic into java anyways so might be easier to test then

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
sumachalla39: I see the intent, but I prefer the explicit logic directly in the controller — it feels clearer for this case. The rule is only used here, and keeping it inline avoids adding extra indirection in the domain model.
Other Controller methods use similar checks but from SessionUtils.
Happy to revisit later if it becomes more widely reused.

jackheslop96: Happy with that rationale 👍

src/main/java/uk/gov/justice/laa/amend/claim/service/AssessmentService.java
```diff
@@ -32,6 +35,9 @@ public class AssessmentService {
     private final Counter assessmentSubmissionFailureCounter;
     private final BigDecimal highValueAssessmentLimit;
 
+    @Value("${feature-flags.is-voiding-enabled: false}")
+    private boolean isVoidEnabled;
```
mboyd-moj: Can you add this to the constructor so we're not using field injection? Should make your tests not need to use reflection too

src/main/java/uk/gov/justice/laa/amend/claim/service/AssessmentService.java
```diff
@@ -107,15 +114,37 @@ public CreateAssessment201Response submitAssessment(ClaimDetails claim, String u
     }
 
     public ClaimDetails getLatestAssessmentByClaim(ClaimDetails claimDetails) {
+        // This will be replaced with filter in data-claims api after VOID api has been implemented
```
mboyd-moj, line 117: Can we make the change to the API now instead of leaving this comment that we might forget about?

sumachalla39: Currently there is no assessment_type in db. After using latest changes, my branch is broken still working on fixing it. I will first ensure it workes with existing api before making any api changes. Please wait before you do further review for this PR

PR #285, opened 2026-03-06
Bc 461

jackheslop96:
## BC-461

[Link to story](https://dsdmoj.atlassian.net/browse/BC-464)

- Security changes for clean bill of health in ZAP report
- `rules.tsv` includes 4 ignored rules with justifications
- Passing job - https://github.com/ministryofjustice/laa-amend-a-claim/actions/runs/22755024082

## Checklist


.zap/rules.tsv
```diff
@@ -1 +1,4 @@
 40045	IGNORE	(Spring4Shell : False positive. We are on a Spring Framework version that is no longer affected by this vulnerability.)
+2	IGNORE	(Private IP Disclosure : False positive. ZAP has spotted something in favicon.svg that looks like an IP address.)
```
mboyd-moj, line 2: This is great 😄

src/main/resources/application.yml
```diff
@@ -71,6 +71,12 @@ management:
     tags:
       application: ${spring.application.name}
 
+server:
+  servlet:
+    session:
+      cookie:
+        same-site: lax
```
mboyd-moj, line 78: what's this?

mboyd-moj (approved):

src/main/resources/application.yml
jackheslop96: ZAP reported the following:
> **Cookie without SameSite Attribute**
> A cookie has been set without the SameSite attribute, which means that the cookie can be sent as a result of a 'cross-site' request. The SameSite attribute is an effective counter measure to cross-site request forgery, cross-site script inclusion, and timing attacks.

Valid values for the SameSite attribute are:
* Strict: The browser sends cookies only for first-party contexts (requests originating from the same site that set the cookie).
* Lax: Cookies are not sent on cross-site sub-requests (e.g., images, AJAX) but are sent when a user navigates to the origin site (e.g., following a link).
* None: Cookies are sent in all contexts, including cross-site (third-party) scenarios.

PR #289, opened 2026-03-06
BC-382: Use feature flag for Void claim button

mboyd-moj:
## What is this PR?

Wires up the feature flag to the Void claim button and removes role check from it so we can see the button before roles are setup

## Checklist

### DIFF
--- a/src/main/java/uk/gov/justice/laa/amend/claim/config/FeatureFlagsConfig.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/config/FeatureFlagsConfig.java
@@ -8,5 +8,5 @@ import org.springframework.context.annotation.Configuration;
 @Configuration
 @ConfigurationProperties(prefix = "feature-flags")
 public class FeatureFlagsConfig {
-    private boolean isVoidingEnabled;
+    private Boolean isVoidingEnabled;
 }
--- a/src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
@@ -17,6 +17,7 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.server.ResponseStatusException;
+import uk.gov.justice.laa.amend.claim.config.FeatureFlagsConfig;
 import uk.gov.justice.laa.amend.claim.models.ClaimDetails;
 import uk.gov.justice.laa.amend.claim.service.AssessmentService;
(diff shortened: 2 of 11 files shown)


src/main/java/uk/gov/justice/laa/amend/claim/config/FeatureFlagsConfig.java
```diff
@@ -8,5 +8,5 @@
 @Configuration
 @ConfigurationProperties(prefix = "feature-flags")
 public class FeatureFlagsConfig {
-    private boolean isVoidingEnabled;
+    private Boolean isVoidingEnabled;
```
mboyd-moj, line 11: this is required because otherwise lombok makes the getter called isVoidingEnabled() and deserialization from the properties file won't work. Doing it this way makes the getter getIsVoidingEnabled() and works nicely with spring

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
```diff
@@ -57,6 +59,10 @@ public String onPageLoad(
         model.addAttribute("claimId", claimId);
         model.addAttribute("submissionId", submissionId);
         model.addAttribute("claim", claim.toViewModel());
+
+        boolean isVoidClaimPresent = featureFlagsConfig.getIsVoidingEnabled() && !claim.isVoided();
+        model.addAttribute("isVoidClaimPresent", isVoidClaimPresent);
```
sumachalla39, line 64: A feature flag doesn’t seem necessary in this case. Because the Void action is strictly role‑based, we can hide the button for unauthorised users and protect the POST API with Spring Security.

src/main/resources/templates/claim-summary.html
```diff
@@ -127,7 +127,7 @@ <h2 class="govuk-summary-card__title" th:text="#{claimSummary.allowedTotalsTable
                                     </button>
                                 </th:block>
 
-                                <th:block sec:authorize="hasRole('ROLE_CLAIM_AMENDMENTS_CASEWORKER')" th:if="${!claim.isVoidClaim}">
+                                <th:block th:if="${isVoidClaimPresent}">
```
sumachalla39, line 130: Is Void action is not role‑based? we can hide the button for unauthorised users

mboyd-moj: I'm leaving this out until https://dsdmoj.atlassian.net/browse/BC-466 so we can play around with the page in the meantime

src/main/java/uk/gov/justice/laa/amend/claim/controllers/ClaimSummaryController.java
mboyd-moj: I don't think it hurts to have both. We're likely to want the user permissions set up ahead of time before turning on the feature so it's useful to be able to restrict this separately so we don't block prod

sumachalla39 (approved):

PR #294, opened 2026-03-10
BC-500 added more validation checks to ensure latest assessment must …

sumachalla39:
Validation to check assessments for VOID Claims and Non-Void claim

## What is this PR?

[Link to story](https://dsdmoj.atlassian.net/browse/BC-500)

## Checklist

### DIFF
--- a/src/main/java/uk/gov/justice/laa/amend/claim/exceptions/InvalidAssessmentException.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/exceptions/InvalidAssessmentException.java
@@ -1,0 +1,8 @@
+package uk.gov.justice.laa.amend.claim.exceptions;
+
+public class InvalidAssessmentException extends RuntimeException {
+
+    public InvalidAssessmentException(String message) {
+        super(message);
+    }
+}
--- a/src/main/java/uk/gov/justice/laa/amend/claim/service/AssessmentService.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/service/AssessmentService.java
@@ -1,16 +1,20 @@
 package uk.gov.justice.laa.amend.claim.service;
 
+import static uk.gov.justice.laa.dstew.payments.claimsdata.model.AssessmentType.ESCAPE_CASE_ASSESSMENT;
+import static uk.gov.justice.laa.dstew.payments.claimsdata.model.ClaimStatus.VOID;
+
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.math.BigDecimal;
+import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.UUID;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Service;
 import uk.gov.justice.laa.amend.claim.client.ClaimsApiClient;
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/justice/laa/amend/claim/exceptions/InvalidAssessmentException.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.laa.amend.claim.exceptions;
+
+public class InvalidAssessmentException extends RuntimeException {
```
mboyd-moj, line 3: I think it'd be fine to just throw RuntimeException and assert the error messages in the tests since we don't have any need to catch these anywhere

src/main/java/uk/gov/justice/laa/amend/claim/service/AssessmentService.java
```diff
@@ -109,37 +113,94 @@ public CreateAssessment201Response submitAssessment(ClaimDetails claim, String u
...
+     * Retrieves the latest assessment for a claim and updates {@link ClaimDetails}
+     * based on VOID or ESCAPE_CASE_ASSESSMENT rules.
+     * VOID: the latest assessment must be VOID, and previous escape‑case values are applied if available.
+     * Non‑VOID: the latest assessment must be ESCAPE_CASE_ASSESSMENT; null types are treated as escape‑case during migration.
+     * @param claimDetails the claim for which assessments should be retrieved
+     * @return the updated {@link ClaimDetails} populated with values from the latest relevant assessment
+     */
     public ClaimDetails getLatestAssessmentByClaim(ClaimDetails claimDetails) {
-        // This will be replaced with filter in data-claims api after VOID api has been implemented BC-500
+        // Fetch the latest 5 records to find the most recent escape case
```
mboyd-moj, line 126: It might be worth getting more opinions from the team, but personally I feel like we shouldn't worry about pagination and just grab all assessments here (using a util method to grab all of the pages behind the scenes if needed). Then we don't need to worry about how many we're fetching when more assessment types start coming back.

src/main/java/uk/gov/justice/laa/amend/claim/exceptions/InvalidAssessmentException.java
sumachalla39: This is a tiny class and beneficial if we want to check logs for invalid assessments of type. Gives you clarity on why its failed before going through details.
image

src/main/java/uk/gov/justice/laa/amend/claim/service/AssessmentService.java
sumachalla39: There’s currently no API that returns all assessments without pagination, and that’s intentional due to performance constraints. During tech standup we noted that, in practice, a claim typically has only 1–2 assessments. To stay safe without introducing unnecessary overhead, I set the page size to 5, which comfortably covers the realistic upper bound while still keeping requests efficient.

Since we only need the latest 2 assessments, there’s no functional reason to fetch all pages. Setting the page size to 5 gives us a comfortable buffer while still avoiding unnecessary calls or iteration logic.

we can revisit it and iterate through all pages, but at this stage the complexity to get all assessments isn’t justified.

Happy to discuss further with the team if this justification isn’t sufficient

mboyd-moj: OK fine with that. Can you add an info log if there are more results to come after the 5 we fetch? That way we know when we need to rethink this

mboyd-moj (approved):

PR #295, opened 2026-03-10
BC-497 Alert when Provider Details API is unavailable during office hours

jackheslop96:
## BC-497

[Link to story](https://dsdmoj.atlassian.net/browse/BC-497)

- Downgraded provider API failure log level from warn to info, as alerting is now handled via Prometheus rather than log-based monitoring
- Added a Prometheus alert (ProviderDetailsApiUnavailable) that fires when the Provider Details API returns errors for 5+ minutes during office hours (Mon–Fri, 07:00–9:30 BST)

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/service/ProviderService.java
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Fetches provider firm name from Provider API
+     *
+     * @param officeAccountNumber the office account number
+     * @return firm name from API or office account number as fallback
+     */
+    public ProviderFirmOfficeDto getProviderFirm(String officeAccountNumber) {
+        LocalTime start = providerApiProperties.getStart().toLocalTime();
+        LocalTime end = providerApiProperties.getEnd().toLocalTime();
+        if (isInTimeRange(LocalTime.now(clock), start, end)) {
```
mboyd-moj: Can we have an info log if it's out of hours

src/main/java/uk/gov/justice/laa/amend/claim/utils/TimeUtils.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public final class TimeUtils {
+
+    public static boolean isInTimeRange(LocalTime now, LocalTime start, LocalTime end) {
+        if (start.equals(end)) {
+            return true;
+        }
+        if (start.isAfter(end)) {
```
mboyd-moj: Ooo nice thinking

src/main/java/uk/gov/justice/laa/amend/claim/utils/TimeUtils.java
```diff
@@ -0,0 +1,18 @@
...
+            return now.isAfter(start) || now.isBefore(end);
+        }
+        return now.isAfter(start) && now.isBefore(end);
```
mboyd-moj: I think we probably want our comparisons to be inclusive. If it's 7.00am I'd say that's in the time range

src/test/java/uk/gov/justice/laa/amend/claim/utils/TimeUtilsTest.java
```diff
@@ -0,0 +1,66 @@
...
+    public void isInTimeRange_whenStartAndEndAreTheSame_returnsTrue() {
+        LocalTime now = LocalTime.of(12, 0);
+        LocalTime start = LocalTime.of(15, 0);
+        LocalTime end = LocalTime.of(15, 0);
+
+        Assertions.assertTrue(TimeUtils.isInTimeRange(now, start, end));
+    }
+
+    @ParameterizedTest
+    @ValueSource(ints = {18, 19, 20, 21, 22, 23, 0, 1, 2, 3, 4, 5, 6, 7, 8})
```
mboyd-moj: I think it'd be fine to just test the edge cases

src/main/java/uk/gov/justice/laa/amend/claim/config/ClockConfig.java
```diff
@@ -0,0 +1,14 @@
...
+import java.time.Clock;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ClockConfig {
+
+    @Bean
+    Clock clock() {
+        return Clock.systemDefaultZone();
```
mboyd-moj: Which timezone is this out of interest?

src/main/java/uk/gov/justice/laa/amend/claim/utils/TimeUtils.java
jackheslop96: `now` will include seconds, nanoseconds. Highly unlikely we will ever get equality here. Also the `LocalTime` library doesn't contain `isAfterOrEqualTo`, `isBeforeOrEqualTo` etc. so it would be a lot of `!isBefore`, `!isAfter`

src/main/java/uk/gov/justice/laa/amend/claim/config/ClockConfig.java
jackheslop96: So technically (in the higher environments) this would be the timezone of the EC2 instance. We can more specifically set this as `Clock.systemUTC()`. I guess we should really align with how the provider details team are setting their OOH schedule.

mboyd-moj: I'd like to keep ours as systemUTC and if we need to have a different timezone for provider team we pass that in config

src/main/java/uk/gov/justice/laa/amend/claim/utils/TimeUtils.java
jackheslop96: So it would be
```
public static boolean isInTimeRange(LocalTime now, LocalTime start, LocalTime end) {
        if (start.equals(end)) {
            return true;
        }
        if (start.isAfter(end)) {
            return !now.isBefore(start) || !now.isAfter(end);
        }
        return !now.isBefore(start) && !now.isAfter(end);
    }
```
or
```
public static boolean isInTimeRange(LocalTime now, LocalTime start, LocalTime end) {
        if (start.equals(end)) {
            return true;
        }
        if (start.isAfter(end)) {
            return !(now.isBefore(start) && now.isAfter(end));
        }
        return !(now.isBefore(start) || now.isAfter(end));
    }
```
personally I think that affects readability but happy to change it if we see fit

mboyd-moj: I think as a utility method this would be better. I know at the moment we're passing in system time and it's almost zero chance we're on the boundary but we could use this for checking something else in the future and forget that we're not inclusive

src/main/java/uk/gov/justice/laa/amend/claim/service/ProviderService.java
```diff
@@ -0,0 +1,48 @@
...
+    private boolean isAvailable() {
+        boolean available;
+        try {
+            HealthDto response = providerApiClient.ping().block();
+            available = response != null && response.getStatus().equals(Status.UP);
+        } catch (Exception e) {
+            available = false;
+        }
+        if (!available) {
+            log.info("Provider details API is unavailable");
```
mboyd-moj: Can we configure an alert for this occurring during office hours? That way we'll at least notice if their API goes down

src/main/java/uk/gov/justice/laa/amend/claim/models/HealthDto.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.laa.amend.claim.models;
+
+import lombok.Data;
+import org.springframework.boot.health.contributor.Status;
+
+@Data
+public class HealthDto {
```
mboyd-moj: Would a record or `@Value` suffice?

jackheslop96: Yes good shout

src/main/java/uk/gov/justice/laa/amend/claim/service/ProviderService.java
jackheslop96: Cool idea yes let me have a look in Sentry to see what can be done

jackheslop96: Discussions had around handling this in Sentry or Pingdom

src/main/java/uk/gov/justice/laa/amend/claim/service/ProviderService.java
```diff
@@ -0,0 +1,48 @@
...
+    private final ProviderApiClient providerApiClient;
+
+    /**
+     * Fetches provider firm office from Provider API
+     *
+     * @param officeAccountNumber the office account number
+     * @return firm office from API or null as fallback
+     */
+    public ProviderFirmOfficeDto getProviderFirm(String officeAccountNumber) {
+        if (isAvailable()) {
```
ahmadsjustice: Is the isAvailable() check adding any value?
Happy path (API is up): We make 2 calls instead of 1 - isAvailable() + the actual provider call.   
Out of hours (API is down): We fail on isAvailable() instead of failing on the actual provider call - same delay either way.
Without isAvailable(), the outcome is identical in both cases - we just skip the extra call on the happy path.

jackheslop96: @ahmadsjustice the key is that we don't want to be creating unnecessary noise with warning logs during out of hours that the lookup failed. But we do want to be alerted in-hours.

Sentry alerts are currently created for any warning log in a given window (think it's 5 minutes). The more alerts we have of this nature, the more likely it is that a genuine alert could be missed.

.helm/laa-amend-a-claim/templates/prometheus-rules.yaml
```diff
@@ -84,4 +84,20 @@ spec:
...
+  - name: application-rules
+    rules:
+    - alert: ProviderDetailsApiUnavailable
+      annotations:
+        message: Provider Details API is returning errors during office hours.
+        dashboard_url: {{ .Values.alerting.grafanaDashboard }}
+      expr: >-
+        (
+          increase(http_client_requests_seconds_count{namespace="{{ .Release.Namespace }}", uri="/api/v1/provider-offices/{officeCode}", outcome!="SUCCESS"}[5m]) > 0
+        )
```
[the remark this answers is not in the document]
ahmadsjustice: interested in knowing each error - message now accurately describes the behaviour.

.helm/laa-amend-a-claim/templates/prometheus-rules.yaml
```diff
@@ -84,4 +84,20 @@ spec:
...
+        and on() (hour() >= 7 and hour() <= 17)
+        and on() (day_of_week() >= 1 and day_of_week() <= 5)
+      for: 0m
+      labels:
```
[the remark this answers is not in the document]
ahmadsjustice: The intent is for this alert to be strict. We want to catch every error, including transient blips. The provider API being unavailable directly impacts the user experience

.helm/laa-amend-a-claim/templates/prometheus-rules.yaml
```diff
@@ -84,4 +84,20 @@ spec:
...
+    rules:
+    - alert: ProviderDetailsApiUnavailable
+      annotations:
+        message: Provider Details API has returned errors in the last 5 minutes during office hours.
+        dashboard_url: {{ .Values.alerting.grafanaDashboard }}
+      expr: >-
+        (
+          increase(http_client_requests_seconds_count{namespace="{{ .Release.Namespace }}", uri="/api/v1/provider-offices/{officeCode}", outcome!="SUCCESS"}[5m]) > 0
+        )
+        and on() (hour() >= 7 and hour() <= 16)
```
BenMillar-MOJ: PDAs opening hours are 07:00  - 21:30

BenMillar-MOJ (changes requested):

BenMillar-MOJ (approved):

PR #321, opened 2026-03-20
BC-522: Add scheduled workflow to cleanup stale ephemeral and e2e environments.

ahmadsjustice:
## BC-522

[Link to story](https://dsdmoj.atlassian.net/browse/BC-522)
Added a new scheduled GitHub Actions workflow that automatically cleans up stale ephemeral and e2e environments older than 7 days. 
Ephemeral environments and e2e test environments can linger in the cluster when PRs go inactive or e2e tests fail (cleanup now only runs on success). This wastes cluster resources. 
The workflow runs daily at midnight UTC and can also be triggered manually.

### DIFF
--- a/.github/workflows/cleanup-stale-environments.yml
+++ b/.github/workflows/cleanup-stale-environments.yml
@@ -1,0 +1,65 @@
+name: Cleanup Stale Environments
+
+on:
+  schedule:
+    - cron: "0 0 * * *" # Daily at midnight UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  cleanup-stale:
+    name: Remove Stale Ephemeral & E2E Environments
+    runs-on: ubuntu-latest
+    environment: dev
+
+    steps:
+      - name: Authenticate to Kubernetes Cluster
+        uses: ministryofjustice/laa-reusable-github-actions/.github/actions/authenticate_to_cluster@768f673017edbb793cf052fa159c272d6c1f9177
+        with:
+          kube-cert: ${{ secrets.KUBE_CERT }}
+          kube-token: ${{ secrets.KUBE_TOKEN }}
+          kube-cluster: ${{ secrets.KUBE_CLUSTER }}
+          kube-namespace: ${{ secrets.KUBE_NAMESPACE }}
+          kube-sa: deploy-user
+
+      - name: Cleanup Stale Environments
+        env:
+          KUBE_NAMESPACE: ${{ secrets.KUBE_NAMESPACE }}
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-stale-environments.yml
```diff
@@ -0,0 +1,65 @@
...
+          kube-token: ${{ secrets.KUBE_TOKEN }}
+          kube-cluster: ${{ secrets.KUBE_CLUSTER }}
+          kube-namespace: ${{ secrets.KUBE_NAMESPACE }}
+          kube-sa: deploy-user
+
+      - name: Cleanup Stale Environments
+        env:
+          KUBE_NAMESPACE: ${{ secrets.KUBE_NAMESPACE }}
+        shell: python3 {0}
+        run: |
```
BenMillar-MOJ, line 32: Are you okay adding this script as a separate file so we don't have a python script in the actions yml file?

BenMillar-MOJ (changes requested):

.github/workflows/cleanup-stale-environments.yml
```diff
@@ -0,0 +1,65 @@
...
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  cleanup-stale:
+    name: Remove Stale Ephemeral & E2E Environments
+    runs-on: ubuntu-latest
+    environment: dev
```
mboyd-moj, line 16: this makes sure that we will only ever run this on dev namespace?

mboyd-moj (approved):

.github/workflows/cleanup-stale-environments.yml
ahmadsjustice: Yes. means the job can only access secrets configured under the dev environment in GitHub. KUBE_NAMESPACE etc all point to dev cluster - won't touch staging or production.

.github/workflows/cleanup-stale-environments.yml
ahmadsjustice: reason for Python over bash is that date comparison and JSON parsing are much cleaner - no jq, no date format workarounds etc. 
Happy to switch it over to bash - we already have inline bash scripts in our pipeline steps.
or we can extract python part to a separate file but prefer inline makes it easy to see exactly what the job does in one place.

BenMillar-MOJ: Fair enough, sounds good to me! 👍

BenMillar-MOJ (approved):

BenMillar-MOJ (approved):
Thank you very much for this!

PR #331, opened 2026-03-25
BC-530: Bulk upload validation

sumachalla39:
## What is this PR?

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+                    Optional.empty(),
+                    Optional.empty(),
+                    Optional.empty(),
+                    Optional.of(AreaOfLaw.LEGAL_HELP),
+                    Optional.of(true),
+                    page,
+                    pageSize,
+                    null);
+        } catch (Exception e) {
+            return null;
```
mboyd-moj: We shouldn't ever swallow an exception, what is the legitimate reason this could fail?

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+                    String ufn = claim.getUniqueFileNumber();
+                    int rowIdx = ufnToRowIdx.get(ufn);
+                    if (matchedUfns.contains(claim.getUniqueFileNumber())) {
+                        errors.add("Line " + (rowIdx + 2) + ": Duplicate Claim found for UFN " + ufn);
+                    } else {
+                        matchedUfns.add(claim.getUniqueFileNumber());
+                    }
+                    BulkUploadCivilClaim row = rows.get(rowIdx);
+                    if (row.getCivilClaimDetails() == null) {
+                        row.setCivilClaimDetails(claimMapper.mapToCivilClaimDetails(claim));
```
mboyd-moj: Copying my comments over from slack:

Here's what I was trying to describe during standup. The emphasis is on improving readability by only having a single level of nesting when looping

```
protected BulkUploadResult mapRowsToClaims(List<T> rows, List<? extends ClaimDetails> claims) {
    var allClaims = getAllClaims(rows);

    var claimsByOfficeCodeAndUfn = allClaims.stream()
            .collect(groupingBy(claim -> Pair.of(claim.getOfficeCode(), claim.getUniqueFileNumber())));

    var errors = new ArrayList<String>();
    for (var row : rows) {
        var claimResponsesForRow =
                claimsByOfficeCodeAndUfn.get(Pair.of(row.getOfficeCode(), row.getUfn()));

        if (claimResponsesForRow == null || claimResponsesForRow.size() != 1) {
            errors.add("You can add your errors in here");
            continue;
        }

        var claimResponseForRow = claimResponsesForRow.getFirst();

        // TODO: This is where you'd do your mapping onto the existing claimDetails object
        CivilClaimDetails existingClaimForRow = null;

        // TODO: This is where you'd map the fields from the CSV onto the existing claimDetails object
        CivilClaimDetails assessedClaimForRow = null;

        // TODO: Can do validation on the claimDetails here, reusing what we already have for single assessment

        claims.add(assessedClaimForRow);
    }

    if (!errors.isEmpty()) {
        return new BulkUploadResult(MAPPING_FAILURE, errors);
    }
    return new BulkUploadResult(SUCCESS, List.of("all is good"));
}

private List<ClaimResponseV2> getAllClaims(List<T> rows) {
    var officeCodes = rows.stream().map(BulkUploadClaim::getOfficeCode).collect(Collectors.toSet());
    return officeCodes.stream()
            .map(this::getAllClaimsForOfficeCode)
            .flatMap(List::stream)
            .collect(Collectors.toList());
}

private List<ClaimResponseV2> getAllClaimsForOfficeCode(String officeCode) {
    // TODO: Get all pages of claims for office code into a single list
    return List.of();
}
```

mboyd-moj: If you look where we're doing row.setCivilClaimDetails(...), this is hidden away inside an iteration over a stream, inside a helper method, inside a loop. I think this is what will make it hard to maintain/add features to. Have you had a look at the rough code I put above? Would be interested to hear your thoughts on that

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+    public static final String SEPERATOR = "-";
+
+    private final ClaimService claimService;
+
+    private final ClaimMapper claimMapper;
+
+    /**
+     * Validates BulkUploadCivilClaim row
+     */
+    private static List<String> validateEachRow(BulkUploadCivilClaim row, int lineNumber) {
```
mboyd-moj: When we discussed this ticket initially we said we would reuse the existing validation we already have by extracting it into one place. Can we do this instead of duplicating here? i.e. construct the new claim object and then run the method to validate it

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+        List<String> errors = new ArrayList<>();
+        String code = row.getOfficeCode();
+        String ufn = row.getUfn();
+
+        if (StringUtils.isBlank(code) || code.length() != 6 || !code.matches(OFFICE_CODE_REGEX)) {
+            errors.add("Line " + lineNumber + ": Invalid office code: " + code);
```
mboyd-moj: We should use "Row" rather than "Line" to match excel's terminology that they're used to

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+                .filter(e -> !matchedUfns.contains(e.getKey()))
+                .forEach(e -> errors.add("Line " + (e.getValue() + 2) + ": UFN "
+                        + e.getKey() + " not found for office code "
+                        + officeCode));
+    }
+
+    /**
+     * Map officeCode with UFN and row index from given rows of csv
+     */
+    public HashMap<String, HashMap<String, Integer>> getOfficeCodeToUfnIdx(
```
mboyd-moj: This can just be Map<String, Map<String, Integer>> as a return type

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+            List<BulkUploadCivilClaim> rows, List<String> errors) {
+        var officeCodeToUfnToRowIdx = new HashMap<String, HashMap<String, Integer>>();
+        Set<String> recordKeys = new HashSet<>();
+
+        for (int i = 0; i < rows.size(); i++) {
+            BulkUploadCivilClaim row = rows.get(i);
+            if (row == null) {
+                continue;
```
mboyd-moj: why would row be null? This feels like an exception

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+            List<String> rowErrors = validateEachRow(row, i + 1);
+
+            if (!rowErrors.isEmpty()) {
+                errors.addAll(rowErrors);
+                continue; // Skip this row for further processing
+            }
+
+            if (code != null && ufn != null) {
+                officeCodeToUfnToRowIdx
+                        .computeIfAbsent(code.trim(), k -> new HashMap<>())
```
mboyd-moj: It feels too late to be doing trimming here, can we trim when we first parse the CSV so we don't need to trim it every time we use it further on?

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+            if (row == null) {
+                continue;
+            }
+            String code = row.getOfficeCode();
+            String ufn = row.getUfn();
+            String key = code + "|" + ufn;
+
+            // --- duplicate CSV row detection by office code and ufn---
+            if (!recordKeys.add(key)) {
+                errors.add("Line " + (i + 2) + ": Duplicate row in CSV for OfficeCode " + code + " and UFN " + ufn);
```
mboyd-moj: Can we add row number to BulkUploadCivilClaim when we first create it so we don't have to keep computing it

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+
+        int totalPages = Optional.ofNullable(first.getTotalPages()).orElse(1);
+
+        // Stream first page
+        Stream<ClaimResponseV2> pageOne = first.getContent().stream();
+
+        // Stream remaining pages
+        Stream<ClaimResponseV2> others = IntStream.range(1, totalPages)
+                .mapToObj(p -> safeFetch(officeCode, p, pageSize))
+                .filter(Objects::nonNull)
```
mboyd-moj: filtering nonNull here seems dangerous, if we don't fetch anything back then something has gone wrong

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+                .filter(r -> r.getContent() != null)
```
mboyd-moj: Can they ever send us no content? Feels like an exception

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+     * Process Claims for each office code and return the matched UFNs from file
+     */
+    public void processAllClaimsForOfficeCodeAndUpdateRow(
+            List<BulkUploadCivilClaim> rows,
+            String officeCode,
+            HashMap<String, Integer> ufnToRowIdx,
+            List<String> errors) {
+        var matchedUfns = new HashSet<String>();
+        int pageSize = 200;
+        fetchClaimsPages(officeCode, pageSize)
```
mboyd-moj: if any of the calls to fetch fail then are we just going to return a 500 error to the user? I think we should be reporting it back as a failure reason

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+            errors.add("Line " + lineNumber + ": Invalid Counsel Costs");
+        }
+        if (OutcomeType.fromCsvLabel(row.getAssessmentOutcome()) == null) {
+            errors.add("Line " + lineNumber + ": Invalid assessment result : " + row.getAssessmentOutcome());
+        }
+
+        if (row.getDisbursements() == null) {
+            errors.add("Line " + lineNumber + ": Invalid Disbursements");
+        }
+        return errors;
```
mboyd-moj: Are we missing validation of the following?
```
    private BigDecimal profitCost;
    private BigDecimal disbursementsVat;
    private BigDecimal totalAllowedVat;
    private BigDecimal totalAllowedInclVat;
```

mboyd-moj: Where is the validation for things like isEscaped?

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+                        + e.getKey() + " not found for office code "
+                        + officeCode));
+    }
+
+    /**
+     * Map officeCode with UFN and row index from given rows of csv
+     */
+    public HashMap<String, HashMap<String, Integer>> getOfficeCodeToUfnIdx(
+            List<BulkUploadCivilClaim> rows, List<String> errors) {
+        var officeCodeToUfnToRowIdx = new HashMap<String, HashMap<String, Integer>>();
```
mboyd-moj: It would be clearer and no more memory to just pass around 
`new HashMap<String, HashMap<String, BulkUploadCivilClaim>>()`
Then you don't need to use what you find to go and lookup the row, you already have it to hand

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+        Set<String> recordKeys = new HashSet<>();
+
+        for (int i = 0; i < rows.size(); i++) {
+            BulkUploadCivilClaim row = rows.get(i);
+            if (row == null) {
+                continue;
+            }
+            String code = row.getOfficeCode();
+            String ufn = row.getUfn();
+            String key = code + "|" + ufn;
```
mboyd-moj: I'd prefer to use one of the Pair classes than constructing our own String key

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+        return claim -> {
+            String ufn = claim.getUniqueFileNumber();
+            return ufn != null && ufnToRowIdx.containsKey(ufn);
+        };
+    }
+
+    private Stream<ClaimResponseV2> fetchClaimsPages(String officeCode, int pageSize) {
+
+        ClaimResultSetV2 first = safeFetch(officeCode, 1, pageSize);
+        if (first == null || first.getContent() == null || first.getContent().isEmpty()) {
```
mboyd-moj: I'm concerned about swallowing these two nulls, they don't feel like regular API behaviour

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,182 @@
...
+import uk.gov.justice.laa.amend.claim.models.AreaOfLaw;
+import uk.gov.justice.laa.amend.claim.models.OutcomeType;
+import uk.gov.justice.laa.amend.claim.service.ClaimService;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.ClaimResponseV2;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.ClaimResultSetV2;
+
+@Component
+@RequiredArgsConstructor
+public class BulkUploadHelper {
+    public static final String SEPERATOR = "-";
```
mboyd-moj: unused

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/civil/BulkUploadCivilClaimCsvMapper.java
mboyd-moj, line 33: Are we missing fields for totalAssessed?

src/main/java/uk/gov/justice/laa/amend/claim/service/BulkUploadCivilService.java
```diff
@@ -1,18 +1,64 @@
...
+            var ufnToRowIdx = officeCodeToUfnToRowIdx.getOrDefault(officeCode, new HashMap<>());
+            // --- Process all claims for this office ---
+            bulkUploadHelper.processAllClaimsForOfficeCodeAndUpdateRow(rows, officeCode, ufnToRowIdx, errors);
+        }
+        if (!errors.isEmpty()) {
+            log.info("Failed validation with {} errors", errors.size());
+            return new BulkUploadResult(VALIDATION_FAILURE, errors);
+        } else {
+            log.info("Successfully validated {} rows", rows.size());
+            String message = "Parsed " + rows.size() + " rows from file";
```
mboyd-moj: Can just log this message instead of constructing similar message twice

mboyd-moj (changes requested):
Quite a few comments on the main code so I've not looked at any of the tests. I feel like we could do with reworking the overarching algorithm used here to make it easier to read/maintain

mboyd-moj:
Can you change the PR title to something more informative so that we end up with a useful commit message once merged?

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
mboyd-moj: It looks like we're missing validation for our amounts, like this:

assessedTotals.assessedTotalVat.error.min=The total VAT must not be negative
assessedTotals.assessedTotalVat.error.max=The total VAT must be less than \u00A31,000,000.00
assessedTotals.assessedTotalVat.error.invalid=The total VAT must be a number with up to 2 decimal places, like 25, 250.00 or 1,250

sumachalla39: Change message with No ecaped civil claim found

build.gradle
```diff
@@ -69,6 +69,7 @@ dependencies {
     implementation("org.apache.commons:commons-csv:1.14.1")
 
     implementation("io.vavr:vavr:0.11.0")
+    implementation("io.netty:netty-codec-http:4.2.12.Final")
```
mboyd-moj: can this be removed? Think pipeline is good now

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/civil/BulkUploadCivilClaim.java
```diff
@@ -1,10 +1,17 @@
...
 import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
-import uk.gov.justice.laa.amend.claim.models.CivilClaimDetails;
+import org.apache.commons.lang3.StringUtils;
 
 @Data
```
mboyd-moj, line 16: I think you can maybe drop this to `@Value` now that we don't need to set civilClaimDetails

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,140 @@
...
+        var officeUfnToClaim = new HashMap<Pair<String, String>, BulkUploadCivilClaim>();
+        Set<Pair<String, String>> appearedKeys = new HashSet<>();
+
+        IntStream.range(0, rows.size()).forEach(i -> {
+            BulkUploadCivilClaim row = rows.get(i);
+            row.setRowNumber(i + ROW_OFFSET); // update row number (CSV header = line 1)
+            String office = row.getOfficeCode();
+            String ufn = row.getUfn();
+
+            Pair<String, String> key = Pair.of(office.trim(), ufn.trim());
```
mboyd-moj: It looks like you're already trimming in the mapping so no need to do it again here

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,140 @@
...
+     */
+    public Map<Pair<String, String>, BulkUploadCivilClaim> getOfficeCodeUfnRows(
+            List<BulkUploadCivilClaim> rows, List<String> errors) {
+
+        var officeUfnToClaim = new HashMap<Pair<String, String>, BulkUploadCivilClaim>();
+        Set<Pair<String, String>> appearedKeys = new HashSet<>();
+
+        IntStream.range(0, rows.size()).forEach(i -> {
+            BulkUploadCivilClaim row = rows.get(i);
+            row.setRowNumber(i + ROW_OFFSET); // update row number (CSV header = line 1)
```
mboyd-moj: Can we set row number in the mapper when we first construct BulkUploadCivilClaim?

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private Stream<ClaimResponseV2> fetchClaimsPages(String officeCode, int pageSize) {
+
+        ClaimResultSetV2 first = safeFetch(officeCode, 1, pageSize);
+        if (first == null || first.getContent() == null) {
+            throw new RuntimeException("No claims found for office code " + officeCode);
+        }
+        if (first.getContent().isEmpty()) {
+            return Stream.empty();
+        }
```
mboyd-moj: I'm just thinking that you could move all of the above into "safeFetch" and then you'll be able to simplify the streaming of remaining pages before by removing the null check

src/main/java/uk/gov/justice/laa/amend/claim/constants/AmendClaimConstants.java
```diff
@@ -53,4 +51,6 @@ public static class Label {
 
     // We allow setting the assessed total values if it's a crime case and has a valid police station fee code
     public static final Set<String> VALID_POLICE_STATION_FEE_CODES = Set.of("INVC");
+    public static final BigDecimal MIN = BigDecimal.ZERO;
+    public static final BigDecimal MAX = BigDecimal.valueOf(1_000_000);
```
mboyd-moj: I think these should be named MIN_CURRENCY and MAX_CURRENCY here as it's not clear what they're for in this context

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,140 @@
...
+            BulkUploadCivilClaim row = rows.get(i);
+            row.setRowNumber(i + ROW_OFFSET); // update row number (CSV header = line 1)
+            String office = row.getOfficeCode();
+            String ufn = row.getUfn();
+
+            Pair<String, String> key = Pair.of(office.trim(), ufn.trim());
+            // Duplicate Rows (officeCode + UFN)
+            if (!appearedKeys.add(key)) {
+                errors.add(String.format(
+                        "Line %d: Duplicate row for office code %s and UFN %s", row.getRowNumber(), office, ufn));
```
mboyd-moj: Change Line -> Row

src/test/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelperTest.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Test
+    void getAllClaimsShouldMatchClaimsAndReturnClaims() {
+        String officeCode = "123456";
+        // Mock claim returned from paging API
+        ClaimResponseV2 claim = new ClaimResponseV2();
+        claim.setUniqueFileNumber("20220101/020244");
+
+        when(claimService.searchClaims(eq(officeCode), any(), any(), any(), any(), any(), eq(1), eq(200), isNull()))
```
mboyd-moj: you should be able to replace this any() with the values we're expecting to go in

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,140 @@
...
+
+        IntStream.range(0, rows.size()).forEach(i -> {
+            BulkUploadCivilClaim row = rows.get(i);
+            row.setRowNumber(i + ROW_OFFSET); // update row number (CSV header = line 1)
+            String office = row.getOfficeCode();
+            String ufn = row.getUfn();
+
+            Pair<String, String> key = Pair.of(office.trim(), ufn.trim());
+            // Duplicate Rows (officeCode + UFN)
+            if (!appearedKeys.add(key)) {
```
mboyd-moj, line 73: Do we need this check anymore now that you're checking for duplicate in the service?

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,140 @@
...
+import uk.gov.justice.laa.amend.claim.models.AreaOfLaw;
+import uk.gov.justice.laa.amend.claim.service.ClaimService;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.ClaimResponseV2;
+import uk.gov.justice.laa.dstew.payments.claimsdata.model.ClaimResultSetV2;
+
+@Component
+@RequiredArgsConstructor
+public class BulkUploadHelper {
+    public static final int ROW_OFFSET = 2;
+    public static final int MAX_ROWS = 1300;
```
mboyd-moj: Where has this come from? Feels like we could handle much higher

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
```diff
@@ -0,0 +1,140 @@
...
+        return claims;
+    }
+
+    public List<ClaimResponseV2> getAllClaims(List<BulkUploadCivilClaim> rows, List<String> errors) {
+        var officeCodes = rows.stream().map(BulkUploadCivilClaim::getOfficeCode).collect(Collectors.toSet());
+        var officeCodeUfnRows = getOfficeCodeUfnRows(rows, errors);
+        return officeCodes.stream()
+                .map(officeCode -> processAllClaimsForOfficeCodeAndUpdateRow(officeCode, officeCodeUfnRows))
+                .flatMap(List::stream)
+                .collect(Collectors.toList());
```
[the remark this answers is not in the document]
mboyd-moj: Happy to keep this as an exception as it's an internal error not user error

src/main/java/uk/gov/justice/laa/amend/claim/bulkupload/BulkUploadHelper.java
sumachalla39: I remember Ben mentioned using around 1300 rows as the limit.
Feel free to adjust this value based on what has been tested and agreed with the team.

src/main/java/uk/gov/justice/laa/amend/claim/service/BulkUploadCivilService.java
```diff
@@ -1,19 +1,128 @@
...
+        if (details.getNetDisbursementAmount() != null) {
+            details.getNetDisbursementAmount().setAssessed(row.getDisbursements());
+        }
+
+        if (details.getCounselsCost() != null) {
+            details.getCounselsCost().setAssessed(row.getCounselCosts());
+        }
+
+        if (details.getAllowedTotalVat() != null) {
+            details.getAllowedTotalVat().setAssessed(row.getTotalAllowedVat());
```
mboyd-moj: Do we need to do something with the vatApplicable flag depending if this is zero or positive?

mboyd-moj (approved):

PR #349, opened 2026-03-31
Sort bulk upload errors

BenMillar-MOJ:
## What is this PR?

- Sort bulk upload errors by storing them as `BulkUploadError`'s which have an optional row number.
- Display these errors in a table of Row Number/ Error Message

Screenshot 2026-03-31 at 16 14 57

## Checklist

### DIFF
--- a/src/main/java/uk/gov/justice/laa/amend/claim/service/BulkUploadService.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/service/BulkUploadService.java
@@ -85,7 +85,10 @@ public abstract class BulkUploadService<T> {
             try {
                 csvHeaderValidator.validate(schemaProvider.getSchema(), parser.getHeaderNames());
             } catch (Exception ex) {
-                errors.add(new BulkUploadError(null, ex.getMessage()));
+                String message = StringUtils.isBlank(ex.getMessage())
+                        ? "Failed to validate CSV header."
+                        : ex.getMessage();
+                errors.add(new BulkUploadError(null, message));
                 return;
             }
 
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/justice/laa/amend/claim/service/BulkUploadService.java
```diff
@@ -110,15 +118,20 @@ protected BulkUploadResult submit(List<? extends ClaimDetails> claimDetails, UUI
...
             }
         }
         var successMessage = String.format("Successfully uploaded %s assessments", claimDetails.size());
         log.info(successMessage);
-        return new BulkUploadResult(SUCCESS, List.of(successMessage));
+        return new BulkUploadResult(SUCCESS, List.of(new BulkUploadError(null, successMessage)));
+    }
+
+    protected static void sortByRowNumber(List<BulkUploadError> errors) {
+        errors.sort(Comparator.comparing(BulkUploadError::rowNumber, Comparator.nullsFirst(Comparator.naturalOrder())));
```
mboyd-moj: Could we use `.stream().sorted(...)` and return the new List so we're not relying on a side effect? Also I think it'd be useful to statically import the Comparator methods to make this easier to read

mboyd-moj: Or consider using a SortedSet to keep the reasons in instead of List so that they're always sorted. Sorted set is still nlog(n) so similar performance to List.sort

mboyd-moj (approved):

PR #381, opened 2026-04-15
Added configuration to docker-compose to allow running app through docker compose alongside other data claims applications

JamieBriggsDev:
## What is this PR?
Currently, there is a parent repo with guidance on running the following apps locally via docker compose:
- https://github.com/ministryofjustice/laa-submit-a-bulk-claim
- https://github.com/ministryofjustice/laa-data-claims-event-service
- https://github.com/ministryofjustice/laa-data-claims-api

This PR adds the setup required to run Amend a Claim alongside the aforementioned apps through the LAA Data Claims Parent Repo:
https://github.com/ministryofjustice/laa-data-claims-parent

## Checklist


PhilDigitalJustice (dismissed):

Dockerfile_internal
```diff
@@ -0,0 +1,38 @@
+# Specify java runtime base image
```
mboyd-moj, line 1: should this file be Dockerfile_internal to match submit or does it serve a different purpose?

docker-compose-app.yml
```diff
@@ -0,0 +1,36 @@
+include:
+  - docker-compose.yml
```
mboyd-moj: I think just add amend-ui to our existing docker-compose, no need for two

docker-compose.yml
```diff
@@ -9,15 +9,15 @@ services:
...
       - -local-response-templating
-  redis:
+  amend-redis:
     image: redis:8-alpine
-    container_name: redis
+    container_name: amend-redis
     restart: always
     ports:
-      - "6379:6379"
+      - "6380:6379" #Updated to 6380 to avoid conflict with SaBC Redis instance
```
mboyd-moj, line 39: can you change the port the app is targeting to 6380 in application.yaml for local profile so it still works for local development

docker-compose-app.yml
```diff
@@ -0,0 +1,36 @@
...
+      FEEDBACK_URL: https://www.smartsurvey.co.uk/s/AaBC_Beta_DevTest
+      LAA_URL: https://dev.your-legal-aid-services.service.justice.gov.uk
+      SENTRY_ENABLED: false # Disabled for docker environment
+      REDIS_URL: redis://amend-redis:6379
+      CLAIMS_API: http://host.docker.internal:8080
+      CLAIMS_TOKEN: f67f968e-b479-4e61-b66e-f57984931e56
+      PROVIDER_API: ${PROVIDER_DETAILS_API_HOSTNAME}
+      PROVIDER_TOKEN: ${PROVIDER_DETAILS_API_ACCESS_TOKEN}
+      SPRING_PROFILES_ACTIVE: local
+      IS_BULK_UPLOAD_ENABLED: true
```
mboyd-moj: Most of these will already be in .env, can we avoid duplicating here?

Dockerfile_internal
JamieBriggsDev: @mboyd-moj Rather, I made submit follow suite on what I did here. Originally on SaBC and other apps, there was one Dockerfile which took in an argument which specified whether to use a pre-made .jar file in the build directory, or build a new one. When I first created this docker-compose solution, I decided it was easier for everyone that it just rebuild whatever is pulled each time to avoid forgetting to run `gradle build` manually everytime. So my idea here was to instead of following this same logic, I made a new file which built from source which can be used for `docker-compose`, and the original one stays for the ECR publish job

mboyd-moj: new file is good, I was just thinking should the name match submit?

JamieBriggsDev: @mboyd-moj My bad, didn't realise I misnamed this one compared to Submit. Will go with internal as that's how the gradle command is being built

mboyd-moj (dismissed):

mboyd-moj (dismissed):

mboyd-moj (approved):

PR #383, opened 2026-04-16
fix: patch Thymeleaf template injection CVEs (3.1.3 → 3.1.4)

ahmadsjustice:
## What is this PR?

- Force Thymeleaf to 3.1.4.RELEASE to close SNYK-JAVA-ORGTHYMELEAF-16078370/72/77/79 (template injection)
- 3.1.4 rejects `@beanName` / `T(...)` in URL attributes and fragment-call arguments
- Replace `${@environment.getProperty(...)}` in `th:href` with a global `${app}` model attribute
-  put `@ThymeleafUtils.xxx()` calls inside fragment args into an outer `th:with` so the fragment only sees a plain variable.


src/main/java/uk/gov/justice/laa/amend/claim/config/AppProperties.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.laa.amend.claim.config;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Data
+@Configuration
+@ConfigurationProperties(prefix = "app")
+public class AppProperties {
```
mboyd-moj: We should be able to use records for ConfigurationProperties these days

mboyd-moj: Should probably be named AppConfig for consistency with our other classes

mboyd-moj (approved):

PR #408, opened 2026-05-06
BC-547: Extract sorting logic into generic classes

mboyd-moj:
## What is this PR?

https://dsdmoj.atlassian.net/browse/BC-547
Decouple sorting for our search page where possible to allow reuse in other pages/apps

## Checklist


src/main/java/uk/gov/justice/laa/amend/claim/controllers/HomePageController.java
```diff
@@ -114,19 +95,34 @@ public String onSubmit(
...
+              .page(1) // If sort is invalid then the page number is no longer meaningful
+              .officeCode(request.getParameter("officeCode"))
+              .submissionDateMonth(request.getParameter("submissionDateMonth"))
+              .submissionDateYear(request.getParameter("submissionDateYear"))
+              .uniqueFileNumber(request.getParameter("uniqueFileNumber"))
+              .caseReferenceNumber(request.getParameter("caseReferenceNumber"));
+      Optional.ofNullable(request.getParameter("page"))
+          .ifPresent(page -> builder.page(Integer.valueOf(page)));
+      Optional.ofNullable(request.getParameter("areaOfLaw"))
+          .ifPresent(areaOfLaw -> builder.areaOfLaw(AreaOfLaw.valueOf(areaOfLaw)));
```
ahmadsjustice, line 121: if we throw IllegalArgumentException NumberException here it will result in 500 error page for user. 
 i-e if areaOfLaw is invalid it will throw IllegalArgumentException 
We can add 
public static Optional<AreaOfLaw> fromString(String value) in AreaOfLaw.java returns Optional if value is valid and here we can do 
AreaOfLaw.fromString(request.getParameter("areaOfLaw")).ifPresent(builder::areaOfLaw);

- Page we can default to 1 and not parse it -  if sort param is wrong we can redirect to page 1. 
- and escapeCase  maybe use BooleanUtils rather then Boolean.valueOf to avoid exception ? 

All this for 1 wrong bookmark

ahmadsjustice (dismissed):

src/main/java/uk/gov/justice/laa/amend/claim/controllers/HomePageController.java
mboyd-moj: I think we can avoid adding the extra handling unless we're actually getting the exceptions through. The sorting one was fixed just because it was spamming our alerts channel

ahmadsjustice (dismissed):

ahmadsjustice (approved):

PR #460, opened 2026-06-22
Add npm to dependabot

mboyd-moj:
## What is this PR?

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -41,3 +41,13 @@ updates:
       github-actions:
         patterns:
           - "**"
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+    groups:
+      npm-dependencies:
+        patterns:
+          - "*"


.github/dependabot.yml
```diff
@@ -41,3 +41,13 @@ updates:
       github-actions:
         patterns:
           - "**"
+  - package-ecosystem: "npm"
```
JamieBriggsDev, line 44: I think alongside this, we should add the `package.json` to track the front end changes and also the workflow to update the webjars also otherwise I don't believe this dependabot change will do anything:
https://github.com/ministryofjustice/laa-submit-a-bulk-claim/blob/main/.github/workflows/frontend-dependencies-update-pipeline.yml

For a bit of context, we added this workflow to SaBC due to WebJars not really updating same time as when MoJ Frontend or GOV UK Updates are published as it's a manual job. This dependabot would work alongside the workflow to check via NPM for those changes, then if the webjar isn't published, it published the webjar so the dependabot for gradle hits it also

JamieBriggsDev (changes requested):
Comment added regarding the context of why we needed this dependabot update on SaBC

.github/dependabot.yml
mboyd-moj: Ah ha. Lets chat this through in tech sync tomorrow as what you're saying makes sense but just want to check we're comfortable with all the extra config for this in amend

PR #475, opened 2026-07-01
Remove inherit secrets from github actions

mboyd-moj:
## What is this PR?

## Checklist

### DIFF
--- a/.github/workflows/feature.yml
+++ b/.github/workflows/feature.yml
@@ -18,7 +18,6 @@ jobs:
     uses: ./.github/workflows/build.yml
     with:
       branch: ${{ github.ref_name }}
-    secrets: inherit
 
   pact-test:
     name: Pact Test
@@ -49,7 +48,6 @@ jobs:
       branch: ${{ github.ref_name }}
       image_tag: ${{ needs.build.outputs.image_tag }}
       aws_region: ${{ vars.ECR_REGION }}
-    secrets: inherit
 
   test:
     name: Test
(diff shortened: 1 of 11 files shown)


.github/workflows/feature.yml
```diff
@@ -18,7 +18,11 @@ jobs:
     uses: ./.github/workflows/build.yml
     with:
       branch: ${{ github.ref_name }}
-    secrets: inherit
+    secrets:
+      PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
```
JamieBriggsDev, line 22: Could we create a short lived token instead of having a PAT Token if we're reorganising this - could also be a seperate PR if required? We will need to get the laa-java-service-deployer APP ID and Private Key added to the repo's variables. Can see how it's used in the old E2E repository but removes the need of having someones PAT token used:
https://github.com/ministryofjustice/bulk-submission-and-fee-scheme-tests-/blob/main/.github/workflows/trigger-e2e-dispatch.yml#L42-L48

JamieBriggsDev (dismissed):
LGTM, left one comment for further improvement

.github/workflows/feature.yml
mboyd-moj: This PR is just a quick fix to unblock the pipelines, can you add that in a ticket?

mboyd-moj: Created https://dsdmoj.atlassian.net/browse/BC-698

BenMillar-MOJ (approved):
Thanks for this!

PR #489, opened 2026-07-16
BC-690: Add enum typeahead inputs for amendable case fields.

ahmadsjustice:
## What is this PR?

[BC-690](https://dsdmoj.atlassian.net/browse/BC-690)

Adds enum/typeahead amendment inputs for Case-tab fields

## Changes

- Added `FieldType.ENUM` with reusable field option support.
- Added an enum/typeahead amendment input partial using the existing autocomplete module.
- Wired selected Case-tab coded fields to render as typeahead inputs.
- Updated unit, view, and E2E coverage for enum/typeahead behaviour.

[BC-690]: https://dsdmoj.atlassian.net/browse/BC-690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/viewfield/ClaimViewField.java
```diff
@@ -15,6 +17,25 @@ public interface ClaimViewField<T extends Claim> {
 
   FieldType getType();
 
+  default List<FieldOption> getOptions() {
+    return List.of();
+  }
+
+  default List<FieldOption> getOptionsIncluding(String selectedValue) {
```
mboyd-moj: I think we probably don't need to support this "including" behaviour. It's going to fail validation if they try to submit it anyway so why prefill it?

src/main/resources/templates/partials/amendments/enum-input.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html lang="en" xmlns:th="http://www.thymeleaf.org">
+<body>
+<th:block th:fragment="amendmentEnumInput(field, labelMessageKey, selectedValue)"
+          th:with="fieldKey=${field.name()},
+                   label=${#messages.msgOrNull(labelMessageKey) ?: fieldKey},
+                   options=${field.getOptionsIncluding(selectedValue)}">
+  <div class="autocomplete-wrapper amend-enum-input">
```
mboyd-moj: Can you configure the typeahead so it doesn't expand the whole table when it expands? e.g. don't want the row growing massive like this, it should overlap

Image

mboyd-moj: Or maybe just chat to Trish about what she wants it to do actually

ahmadsjustice: changed to not expand the table row , sent message to Trish for review

src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/viewfield/StandardFeeCategory.java
```diff
@@ -0,0 +1,32 @@
+package uk.gov.justice.laa.amend.claim.viewmodels.viewfield;
+
+public enum StandardFeeCategory implements FieldOption {
+  ONE_A("1A"),
```
mboyd-moj: This looks iffy, how about CATEGORY_1A?

src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/viewfield/DesignatedAccreditedRepresentative.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.justice.laa.amend.claim.viewmodels.viewfield;
+
+public enum DesignatedAccreditedRepresentative implements FieldOption {
+  VALUE_1("1"),
```
mboyd-moj: The naming here seems weird, is it highlighting that we're using an enum where we should just be using an int?

src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/viewfield/StandardFeeCategory.java
ahmadsjustice: used  `ONE_A`  to avoid repeating the enum type in the constant name. 
Since the fully qualified usage is  `StandardFeeCategory.ONE_A `, adding  `CATEGORY_ ` feels a bit redundant ( `StandardFeeCategory.CATEGORY_1A` ).
Happy to rename it.

src/main/java/uk/gov/justice/laa/amend/claim/viewmodels/viewfield/DesignatedAccreditedRepresentative.java
ahmadsjustice: The data dictionary only lists  1,2,3,4,5  i-e we don’t currently have meaningful labels for these codes.
If there are domain names for these values, I can rename the constants to those.
better represented as simple  FieldOption  values rather than a dedicated enum ?

mboyd-moj (approved):

PR #507, opened 2026-07-27
Makes the autocomplete alphabetical when typing

JamieBriggsDev:
## What is this PR?

[Link to story](https://dsdmoj.atlassian.net/browse/BC-701)

Should make it so when the user types into an autocomplete box, the options are ordered based on the left of the values and not if there is a string match anywhere in the values.

Screenshot 2026-07-28 at 17 37 55

## Checklist

### DIFF
--- a/src/main/resources/sass/styles.scss
+++ b/src/main/resources/sass/styles.scss
@@ -60,6 +60,13 @@
   display: none; /* Hidden until required */
 }
 
+/* Without this, the wrapper doesn't form its own stacking context, so the
+   dropdown arrow's negative z-index escapes to the document root and gets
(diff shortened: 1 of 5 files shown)


src/main/resources/sass/styles.scss
```diff
@@ -60,6 +60,13 @@
   display: none; /* Hidden until required */
 }
 
+/* Without this, the wrapper doesn't form its own stacking context, so the
+   dropdown arrow's negative z-index escapes to the document root and gets
+   painted behind the page background instead of behind just the input. */
```
mboyd-moj, line 65: :mindblown:

mboyd-moj (approved):

PR #520, opened 2026-07-29
Fix/null vs empty amendment comparison

ahmadsjustice:
- Fix spurious "Not applicable" entries appearing in the amended column for boolean and number fields the user never changed.

### DIFF
--- a/src/main/java/uk/gov/justice/laa/amend/claim/forms/amendments/AmendmentForm.java
+++ b/src/main/java/uk/gov/justice/laa/amend/claim/forms/amendments/AmendmentForm.java
@@ -98,15 +98,15 @@ public class AmendmentForm {
     var original = originalForm.getInputs().get(key);
     var current = inputs.get(key);
 
-    if (original == null && current == null) {
+    if (isBlank(original) && isBlank(current)) {
       return false;
     }
 
-    if (original == null || current == null) {
+    if (isBlank(original) || isBlank(current)) {
       return true;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/justice/laa/amend/claim/forms/amendments/AmendmentForm.java
```diff
@@ -98,15 +98,15 @@ public boolean isAmendment(String key, AmendmentForm originalForm) {
     var original = originalForm.getInputs().get(key);
     var current = inputs.get(key);
 
-    if (original == null && current == null) {
+    if (isBlank(original) && isBlank(current)) {
```
mboyd-moj, line 101: How come these are blank in the first place? I thought they should always be null because of StringConverter

mboyd-moj: Happy for this defensive code to go in, but I'd also like us to make sure we're always using null rather than empty string so we're consistent

mboyd-moj (approved):

src/main/java/uk/gov/justice/laa/amend/claim/forms/amendments/AmendmentForm.java
ahmadsjustice: formatBooleanValue & formatNumberValue were converting null to empty. made the changes

PR #586, opened 2026-08-20
test(BC-724): Move e2e module to top level

mboyd-moj:
## What is this PR?

This is step 1/3
1. Move e2e module to top level
2. Move all other src into a ui module
3. Change e2e pipeline to match submit's

## Checklist


JamieBriggsDev (approved):
LGTM - My heart dropped when I seen 108 files, but clearly it's just a lot of files moved and I can see the pipelines still work actions so I'm happy. Also double checked the files which have actually been modified.

PR #587, opened 2026-08-20
test(BC-724): Move src into a ui module

mboyd-moj:
## What is this PR?

This is step 2/3
[~~1. Move e2e module to top level~~](https://github.com/ministryofjustice/laa-amend-a-claim/pull/586)
1b. (Included here) create an e2e src directory
2. Move all other src into a ui module
3. Change e2e pipeline to match submit's

## Checklist


JamieBriggsDev (changes requested):
Could you check `gulpfile.ts` as well as I would imagine this may need some fixes also. Didn't see it here

mboyd-moj:
@JamieBriggsDev It's moved into ui module so paths should not need to change

JamieBriggsDev (approved):
LGTM

PR #589, opened 2026-08-21
test(BC-724): Refactor e2e pipeline to match submit

mboyd-moj:
## What is this PR?

## Checklist


.github/workflows/main.yml
```diff
@@ -25,23 +25,31 @@ jobs:
...
-      DATA_CLAIMS_API_IMAGE: ${{ secrets.DATA_CLAIMS_API_IMAGE }}
-      E2E_DB_USER: ${{ secrets.E2E_DB_USER }}
-      E2E_DB_PASSWORD: ${{ secrets.E2E_DB_PASSWORD }}
+      pat_token: ${{ secrets.PAT_TOKEN }}
+
+  e2e-tests:
+    name: E2E Tests
+    uses: ./.github/workflows/e2e-tests.yml
+    secrets:
+      pat_token: ${{ secrets.PAT_TOKEN }}
```
JamieBriggsDev, line 39: One day it would be good to get rid of this PAT token

.github/workflows/snyk-app-scan.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           client_secret: ${{ secrets.SNYK_CLIENT_SECRET }}
 
       - name: Run Snyk app vulnerability scan
-        run: snyk test --severity-threshold=low --sarif-file-output=snyk-app.sarif
+        run: snyk test laa-amend-a-claim-ui/ --severity-threshold=low --sarif-file-output=snyk-app.sarif
```
JamieBriggsDev, line 43: Would this mean that our E2E module would never get snyk scanning? I think it's okay given our E2E isn't a service we're deploying anywhere, but perhaps it should be clear in a comment next to this

JamieBriggsDev (changes requested):
Added couple comments, love to see this getting more closely aligned

.github/workflows/snyk-app-scan.yml
mboyd-moj: Yeah this is same as in the common pipeline I believe, didn't realise this was the case! I think it's actually already what we had when e2e was hidden within src

mboyd-moj: I'm giving "snyk test --all-projects" a go. Lets see if the e2e module had any vulns 😬

.github/workflows/e2e-tests.yml
```diff
@@ -1,335 +1,201 @@
...
-          check "/css/app.min.css"
-          
-          echo "Running E2E tests..."
-          ../gradlew :e2e:clean :e2e:test --no-daemon
-
-      - name: Generate Allure Report
+        working-directory: laa-data-claims-parent/laa-amend-a-claim
+        run: ./gradlew :e2e:test --no-configuration-cache
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ahmadsjustice, line 155: should this be secrets.pat_token ?

.github/scripts/wait-for-services.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Waits for actuator health to be UP for submit and data claims services.
```
ahmadsjustice: nit: amend and data claims services

.github/workflows/e2e-tests.yml
mboyd-moj: For running the tests we can just use GITHUB_TOKEN. It's only when we build the app and it's dependencies that we need PAT as this will pull in the private repo PDA

.github/workflows/snyk-app-scan.yml
mboyd-moj: It does. I'll do this separately to this PR though as it needs some investigation

